PR #21, opened 2019-06-12
Versioned example deployment yamls

kermat:
Specifying tags for each of the container images in the LINSTOR
deployment example yaml will prevent Kubernetes from pulling down
incompatible versions when "latest" gets updated.

linstor-csi-1.14.yaml: tagged images that work with k8s 1.14.x
linstor-csi-dev.yaml:  original which always pulls latest


haySwim:
:shipit:

PR #26, opened 2019-07-02
Add balancer

PiotrProkop:
Adding new field `placementPolicy` when set to `balance` or `balanced`. CSI will create Volume in the same `failure-domain` and pick `node/storagePool/prefNic(numa node)` based on utilization


pkg/balancer/balancer.go
```diff
@@ -0,0 +1,274 @@
...
+	NodeName        string
+}
+
+type NodeLinstorClient interface {
+	GetStoragePools(ctx context.Context, nodeName string, opts ...*lapi.ListOpts) ([]lapi.StoragePool, error)
+}
+
+func GetInternalk8sClient() (clientset kubernetes.Interface, err error) {
+
+	// setup k7s client
```
haySwim: Surely, this is a k8s client :)

pkg/client/linstor.go
```diff
@@ -228,6 +244,15 @@ const (
 	localStoragePolicyIgnore
 )
 
+type placementPolicy int
```
haySwim: I like the  `placementPolicy` name a lot, but I don't like it in conjunction with `localStoragePolicy`.

To me there are two things at play here: where the volumes are physically created and where they're reachable from after they've been created. Right now, `localStoragePolicy` policies conflate those to two things and `placementPolicy` effects the phsyical placement, so it's an overlap. I think our choices are then to split these up, or conflate under one thing.

Either way, the number of behaviours is going to be (number of local storage placement policies) x (number of network access policies).

I think, when broken up in this way, there are only two network access policies: you can either allow it (if the volumes are even capable) or not. Therefore, I think we should have `placementPolicy` policies solely affect linstor's behavior on where the volumes are physically assigned and then have a boolean field like `allowRemoteVolumeAccess` that determines if we can count that towards CSI's understanding of volume accessibility. That still comes under the constraint of the `localStoragePool`, so we have to account for that.

`localStoragePolicy` would just go away if we did this.

So, including this PR for placements we have:
automatic (let autoplace do it)
manual (list of node names)
balance (this very PR)
followTopology (put volumes along topology preferences. I don't like this name, so consider it a placeholder)

Right now Automatic is the default and Manually is implicit through providing a nodeList e.g, it overrides other options. so we can set the appropriate policy when we parse the params.

So our two explicit placement policies are `balance` and `followTopology`

So, if I figured that out correctly, that's 4 local storage placement policies and 2 remote access policies, giving us 8 different (theoretical) behaviours all handled by the `Create` and `AccessibleTopologies` functions, which is really too much to keep track of. So these need to be split out into separate functions. Probably two functions for each placement policy, each of which has access to the params, so it can account for remote access policy and their other quirks.

So at this point, I think having an interface for these is reasonable. Something like:

```go
type volumeScheduler interface {
	Create(ctx context.Context, vol *volume.Info, req *csi.CreateVolumeRequest)  error
        AccessibleTopologies  (ctx context.Context, vol *volume.Info)  ([]*csi.Topology, error)
}
```

and have the appropriate volumeScheduler take over after this line: https://github.com/LINBIT/linstor-csi/blob/master/pkg/client/linstor.go#L577

So the `parsePlacementPolicy` would instead be `newVolumeScheduler` and return a `volumeScheduler` and not a `placementPolicy`

So... if you think that is reasonable, what I'd like from you is.

Take the balancer Create and accessibility logic out of `pkg/client/linstor.go` and implement that `volumeScheduler` interface in your `pkg/balancer/balancer.go` file. With the assumption that there is a boolean `allowRemoteVolumeAccess` parameter which determines whether volumes can be accessed over the network. 

Sorry if this is a lot of changes, but you happened to hit a spot in the code that I vowed to split up if any more policies got introduced. The upside is that the creation logic should be vastly simpler to understand.

PiotrProkop: I totally agree :smile:

PiotrProkop:
@haySwim I have implemented the `volumeScheduler` interface, please let me know if it is correct.
There are couple of `TODOs` mostly me reusing your non-exported functions so prolly we should move it to some `utils` package

PiotrProkop:
@deleteriousEffect Hi, I have modified this PR to work with new interface :smile:

pkg/topology/scheduler/balancer/balancer.go
```diff
@@ -0,0 +1,420 @@
...
+
+	params, err := volume.NewParameters(vol.Parameters)
+	if err != nil {
+		return fmt.Errorf("unable to create volume due to bad parameters %+v: %v", vol.Parameters, err)
+	}
+
+	if !params.AllowRemoteVolumeAccess {
+		return fmt.Errorf("placementPolicyBalance cannot work on on local storage")
+	}
+
```
deleteriousEffect: It might be good to add a guard clause if the `params.StoragePool != ""` since this scheduler picks a storage pool automatically and if the user has specified a storage pool it might indicate they expect different behavior.

PiotrProkop: :+1:

deleteriousEffect:
lgtm

PR #35, opened 2019-08-22
Allow to reuse mounted volume

kvaps:
fixes: #34

PR #37, opened 2019-09-25
Add example for kubernetes 1.16

kvaps:
Old apiVersions removed since kubernetes 1.16.0.
Example manifest is not working anymore.

### DIFF
--- a/examples/k8s/deploy/linstor-csi-1.16.yaml
+++ b/examples/k8s/deploy/linstor-csi-1.16.yaml
@@ -1,0 +1,379 @@
+---
+kind: StatefulSet
+apiVersion: apps/v1
+metadata:
+  name: linstor-csi-controller
+  namespace: kube-system
+spec:
+  serviceName: "linstor-csi"
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: linstor-csi-controller
+        role: linstor-csi
+    spec:
+      priorityClassName: system-cluster-critical
+      serviceAccount: linstor-csi-controller-sa
+      hostNetwork: true
+      containers:
+        - name: csi-provisioner
+          image: quay.io/k8scsi/csi-provisioner:v1.2.0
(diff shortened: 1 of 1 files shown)


deleteriousEffect:
Thanks for staying on top of the new k8s releases!

Rather than changing the yaml for k8s 1.14, could you create a new yaml for 1.16 i.e., ` examples/k8s/deploy/linstor-csi-1.16.yaml` with your changes?

We should be releasing more versioned yamls soon, so this convention should be more obvious.

kvaps:
@deleteriousEffect [`StatefulSet apps/v1`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.14/#statefulset-v1-apps) and [`DaemonSet apps/v1`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.14/#daemonset-v1-apps) are exist and working even since `v1.9` release. Should I do that?

deleteriousEffect:
I wouldn't worry about it, unless there's an issue that changing the api on
those versions of k8s would resolve.

People are always welcome (and encouraged) to modify the example manifest
file to fit their need.

kvaps:
OK, job is done.

@deleteriousEffect could we make new release `v0.7.3` and bump linstor-csi images also?

deleteriousEffect:
Thanks!

I don't think a version bump is appropriate since the plugin itself hasn't changed. Nothing in the binary or the dockerfile would behave meaningfully differently and that's fundamentally what's being versioned.

kvaps:
#35 is not released yet, and manifests still use old v0.7.0 image

deleteriousEffect:
You are right about #35, sorry! Please give me a bit of time on that as there are a few other things in the pipeline that take priority,

There's a patch forthcoming with image updates and other changes to the manifests soon.

kvaps:
No problem, thank you! 👍

PR #44, opened 2019-11-15
Implement csi expand volume

liuxin95:
**What would you like to be added**
    For some reason we need for capacity expansion function in our business, but I found that linstor csi driver did not implement the capacity expansion function

**How did you resolve it**
    In the linstor document, I found that linstor provided expansion function, but it was not implemented in linstor csi
image
and the expansion interface was implemented in [golinstor](https://github.com/LINBIT/golinstor)
image
image
 that is to say, I only needed to implement ControllerExpand and NodeExpand in csi by calling the expansion interface in golinstor.

**How do you test**
- The test env
```
1. kubernete version ==> 1.14.7
2. kubelet ==> --feature-gates=ExpandCSIVolumes=true
3. StorageClass ==> allowVolumeExpansion: true
```
- I created a 40G PVC and mounted it in a pod
image
image

- then I resized the PVC to 50G,  Kubernetes resize controller detects changes in PVC and first calls the ControllerExpand interface to expand volume
image
image

- kubelet call NodeExpand to expand fs
image
image

ok. Everything looks fine

### DIFF
--- a/pkg/client/linstor.go
+++ b/pkg/client/linstor.go
@@ -26,6 +26,8 @@ import (
 	"io"
 	"regexp"
 	"strings"
+	"os/exec"
+	"log"
 
 	lapi "github.com/LINBIT/golinstor/client"
 	"github.com/LINBIT/linstor-csi/pkg/linstor"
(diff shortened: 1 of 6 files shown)


w00jay:
Thank you for the PR.  Just wanted to give you a heads up that review will take a bit as the KubeCon is  next week.

liuxin95:
Anybody notice this PR?

w00jay:
I apologize for the delay.  I will test it this week.

liuxin95:
@w00jay  That would be great. Thank you

rck:
@liuxin95 Thank you for the detailed explanation. For the time being, I will take care of the PR. It is on top of my TODO list, I will provide feedback soon. Sorry, that this takes a bit longer than one would expect.

pkg/client/linstor.go
```diff
@@ -990,3 +992,128 @@ func nil404(e error) error {
...
+	default:
+		return fmt.Errorf("volume %s format is %s, unsupported expand currently",
+			source, fstype)
+	}
+	if err != nil {
+		return fmt.Errorf("resize filesystem failed, device: %s, path: %s, err: %v", source, target, err)
+	}
+
+	return nil
+}
```
rck: This and its helper functions should not be handrolled. Please use [existing k8s infrastructure](https://godoc.org/k8s.io/kubernetes/pkg/util/resizefs)

pkg/driver/driver.go
```diff
@@ -478,7 +492,7 @@ func (d Driver) ControllerUnpublishVolume(ctx context.Context, req *csi.Controll
 	vol, err := d.Storage.GetByID(ctx, req.VolumeId)
 	if err != nil {
 		return nil, status.Errorf(codes.Internal,
-			"ControllerUnpublishVolume failed for %s: %v", req.GetVolumeId(), err)
+			"ControllerUnpublishVolume failed for %s: %v",req.GetVolumeId(), err)
```
rck: This is just one example, but when I open and close files, it looks like you did not use gofmt (or and IDE that formats Go code correctly). Please do so, otherwise this produces diffs on follow up commits. `gofmt` is the canonical way.

pkg/client/linstor.go
```diff
@@ -990,3 +992,128 @@ func nil404(e error) error {
...
+	args := []string{"-d", deviceMountPath}
+	output, err := exec.Command("xfs_growfs", args...).CombinedOutput()
+
+	if err == nil {
+		logrus.Infof("device %s resized successfully", deviceMountPath)
+		return true, nil
+	}
+
+	resizeError := fmt.Errorf("resize of device %s failed: %v. xfs_growfs output: %s", deviceMountPath, err, string(output))
+	return false, resizeError
```
rck: Very minor, and this code should be removed anyways (because of existing functionality in k8s), but please always handle the error case and early exit there, but not vice versa.
```go
x, err := foo()
if err != nil {
        return err
}
...
return nil
```

pkg/volume/volume.go
```diff
@@ -452,3 +452,8 @@ type Mounter interface {
 	Mount(vol *Info, source, target, fsType string, options []string) error
 	Unmount(target string) error
 }
+
+type Expand interface {
+	NodeExpand(source, target string) error
+	ControllerExpand(ctx context.Context, vol *Info) error
+}
```
rck: Typically Go interfaces are named with "er" at the end. See the "Mount**er**" as an example. Please follow that and rename to `Expander`.

kvaps:
Just wanted to link this PR to issue https://github.com/kubernetes/enhancements/issues/556#issuecomment-573547228

kvaps:
Hi, any update on this?
Other plugins eg. ceph-csi already supporting resizing feature, it is working since 1.16 release.

A good article about this:
https://kubernetes.io/blog/2018/07/12/resizing-persistent-volumes-using-kubernetes/

rck:
A version based on this was merged with #70. Thanks.

PR #53, opened 2020-01-27
Implement NodeGetVolumeStats

lalinsky:
I need this to monitor filesystem usage of Linstor-backed PVs. While I could use other monitoring tools, it would be great to have it directly integrated into K8s.


rck:
Thank you! I'm currently a bit busy with other things, but I hope I can get to it end of this week.

rck:
Thanks, this got merged with 4ccc6e395 . I just added a bit of cleanup to the unconditional mkdir/rmdir mock code. Better be safe than sorry :)

PR #55, opened 2020-02-02
Disable host network

kvaps:
Hi, linstor-csi plugin is working fine without host networking.
This parameter was given here by mistake, actually I'm not using hostNetwork in [kube-linstor](https://github.com/kvaps/kube-linstor/blob/master/helm/kube-linstor/templates/linstor-csi.yaml) project

### DIFF
--- a/examples/k8s/deploy/linstor-csi-1.16.yaml
+++ b/examples/k8s/deploy/linstor-csi-1.16.yaml
@@ -19,8 +19,6 @@ spec:
     spec:
       priorityClassName: system-cluster-critical
       serviceAccount: linstor-csi-controller-sa
-      hostNetwork: true
-      dnsPolicy: ClusterFirstWithHostNet
       containers:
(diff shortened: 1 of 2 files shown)


w00jay:
The prior art establishes the hostnetworked pod to correctly resolve K8s networking.  It is 'required' [here](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/) - 

>For Pods running with hostNetwork, you should explicitly set its DNS policy “ClusterFirstWithHostNet”.

kvaps:
But is there any specific reason for using hostNetowking for csi-pods?

w00jay:
Thanks for question.  This made me realize it's not required in the CSI as you mentioned originally in the PR.  It is used in the operator and the helm charts as LINSTOR container itself needs to be in the same network namespace as the kernel to properly function, but none of that matters here.  8-) .

Issue #54, opened 2020-02-02
Volumes are creating with nodeAffinity set


kvaps:
When you having hundreds of PV it might create overload the api-server and provisioner, also there is no opportunity to using PVs on newly created nodes afterwards, example:

```yaml
apiVersion: v1
kind: PersistentVolume
metadata:
  annotations:
    pv.kubernetes.io/provisioned-by: linstor.csi.linbit.com
  creationTimestamp: "2020-01-29T16:21:58Z"
  finalizers:
  - kubernetes.io/pv-protection
  - external-attacher/linstor-csi-linbit-com
  name: pvc-ffc3e321-8f4a-4036-ae67-0ed27b27fb96
  resourceVersion: "2510909725"
  selfLink: /api/v1/persistentvolumes/pvc-ffc3e321-8f4a-4036-ae67-0ed27b27fb96
  uid: 98e518b9-0277-442d-bc75-e1973e407740
spec:
  accessModes:
  - ReadWriteOnce
  capacity:
    storage: 150Gi
  claimRef:
    apiVersion: v1
    kind: PersistentVolumeClaim
    name: vol-data-mariadb-hc1-md14-1-0
    namespace: hosting
    resourceVersion: "2510907409"
    uid: ffc3e321-8f4a-4036-ae67-0ed27b27fb96
  csi:
    driver: linstor.csi.linbit.com
    fsType: ext4
    volumeAttributes:
      storage.kubernetes.io/csiProvisionerIdentity: 1576375528840-8081-linstor.csi.linbit.com
    volumeHandle: pvc-ffc3e321-8f4a-4036-ae67-0ed27b27fb96
  mountOptions:
  - errors=remount-ro
  nodeAffinity:
    required:
      nodeSelectorTerms:
      - matchExpressions:
        - key: linbit.com/hostname
          operator: In
          values:
          - m16c14
      - matchExpressions:
        - key: linbit.com/hostname
          operator: In
          values:
          - m9c9
      - matchExpressions:
        - key: linbit.com/hostname
          operator: In
          values:
          - m10c1
      - matchExpressions:
        - key: linbit.com/hostname
          operator: In
          values:
          - m10c10
      - matchExpressions:
        - key: linbit.com/hostname
          operator: In
          values:
          - m10c11
      - matchExpressions:
        - key: linbit.com/hostname
          operator: In
          values:
          - m10c12
      - matchExpressions:
        - key: linbit.com/hostname
          operator: In
          values:
[...]

PR #56, opened 2020-02-02
Disable topology feature

kvaps:
- Assigned nodeAffinity cannot be changed for existing PVs over time, that's blocking usage of them in newly created nodes. (see: https://github.com/kubernetes/kubernetes/issues/69528)
- It also heap up each PV resource if you having many nodes cluster. (see: #54)
- This feature is disabled by default for external csi-provisioner 
  (see: https://kubernetes-csi.github.io/docs/topology.html)

fixes https://github.com/LINBIT/linstor-csi/issues/54

### DIFF
--- a/examples/k8s/deploy/linstor-csi-1.14.yaml
+++ b/examples/k8s/deploy/linstor-csi-1.14.yaml
@@ -34,10 +34,10 @@ spec:
         - name: csi-provisioner
           image: quay.io/k8scsi/csi-provisioner:v1.3.0
           args:
-            - "--csi-address=$(ADDRESS)"
             - "--v=5"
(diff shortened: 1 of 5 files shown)


w00jay:
Thank you for the PR.  This is indeed an idea we are interested, but I will have to do some research first.  Thank you for your patience.

WanzenBug:
I think we can close this PR and associated issue?

* csi-provisioner has this enabled by default now
* we have the [linstor-affinity-controller](https://github.com/piraeusdatastore/linstor-affinity-controller) should the affinity change at a later time
* we have the `allowRemoteVolumeAccess` option, which acts as a filter on how specific the set affinity is

PR #57, opened 2020-02-02
Update csi images to latest versions for 1.17

kvaps:
There is also one breaking change about snapshots
https://github.com/kubernetes-csi/external-provisioner/blob/master/CHANGELOG-1.5.md

### DIFF
--- a/examples/k8s/deploy/linstor-csi-1.17.yaml
+++ b/examples/k8s/deploy/linstor-csi-1.17.yaml
@@ -22,7 +22,7 @@ spec:
       hostNetwork: true
       containers:
         - name: csi-provisioner
-          image: quay.io/k8scsi/csi-provisioner:v1.4.0
+          image: quay.io/k8scsi/csi-provisioner:v1.5.0
           args:
             - "--csi-address=$(ADDRESS)"
             - "--v=5"
@@ -36,7 +36,7 @@ spec:
             - name: socket-dir
               mountPath: /var/lib/csi/sockets/pluginproxy/
         - name: csi-attacher
-          image: quay.io/k8scsi/csi-attacher:v2.0.0
+          image: quay.io/k8scsi/csi-attacher:v2.1.1
           args:
             - "--v=5"
             - "--csi-address=$(ADDRESS)"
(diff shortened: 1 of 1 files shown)


examples/k8s/deploy/linstor-csi-1.17.yaml
```diff
@@ -22,7 +22,7 @@ spec:
       hostNetwork: true
       containers:
         - name: csi-provisioner
-          image: quay.io/k8scsi/csi-provisioner:v1.4.0
+          image: quay.io/k8scsi/csi-provisioner:v0.5.0
```
w00jay: Hi, thank you for the PR.  Surely you meant v1.5.0 here, though?

kvaps: My bad, fixed!

kvaps:
OK, Image updated, which params do you mean?

w00jay:
Hi, thank you for the update.  The entire block for the snapshotter should now read;

name: csi-snapshotter
image: quay.io/k8scsi/csi-snapshotter:v2.0.1
args:
- "-timeout=4m"
- "-csi-address=$(ADDRESS)"

which updates the two parameters' names and format.  Thank you so much.

examples/k8s/deploy/linstor-csi-1.17.yaml
```diff
@@ -49,7 +49,7 @@ spec:
             - name: socket-dir
               mountPath: /var/lib/csi/sockets/pluginproxy/
         - name: csi-snapshotter
-          image: quay.io/k8scsi/csi-snapshotter:v1.2.2
+          image: quay.io/k8scsi/csi-snapshotter:v2.0.1
           args:
             - "--csi-address=$(ADDRESS)"
             - "--timeout=120s"
```
kvaps: @w00jay, like just put `--timeout=4m` here?
Excuse me. I little confused, shouldn't we also update timeout option for other images as well?
BTW, you can always do that in separated commit

w00jay: It seems the parameter's changed only for the snapshotter for the moment.  I will follow up.  Thank you!

PR #61, opened 2020-02-21
Do not use DfltDisklessStorPool in balancer scheduler

PiotrProkop:
Linstor v1.4.2 introduced that DfltDisklessStorPool is not filtered which is breaking balancer scheduler. This is fix for that.

### DIFF
--- a/pkg/topology/scheduler/balancer/balancer.go
+++ b/pkg/topology/scheduler/balancer/balancer.go
@@ -129,6 +129,10 @@ func getNodesUtil(ctx context.Context, nClient NodeLinstorClient, selectedNodes 
 		}
 
 		for _, sp := range spls {
+			if sp.ProviderKind == lapi.DISKLESS {
+				continue
+			}
+
 			nicName := sp.Props[PrefNicPropKey]
 			free := sp.FreeCapacity
 			total := sp.TotalCapacity


rck:
Please do not mix unrelated changes in one commit. The change to the `Dockerfile` is unrelated. And it actually breaks LINBIT's build system. There is a reason things are the way they are. So please just refactor the PR to *not* include the change to the `Dockerfile` and force push your branch.

Thanks, rck

PiotrProkop:
@rck you are right, I shouldn't have mixed 2 changes in one commit, I just couldn't use this Dockerfile to build image even from `master` branch. I fixed that, thanks for review!

rck:
Hi Piotr,

guess we are getting there, but I think this is a rather weird/wrong fix. I don't know the balancer code to be honest, but just from the PR message (sorry, you provided ways to less information why you do what you do), which is "Do not use DfltDisklessStorPool in balancer scheduler", filtering all `!= LVM` looks wrong. This - while probably not important in you particular case - breaks ZFS. If you really would want to check for the "DfltDisklessStorPool", you would filter by "DefaultDisklessStoragePoolName" (there is a constant for that). And then still, I'd assume that this is not even what you are trying to accomplish, because (again, I have to guess), I'd assume you want to filter all diskless pools (in general there could be diskless pools manually created). If you want to filter all diskless pools, you would want to filter by `sp.ProviderKind != lapi.DISKLESS`

In the future please provide more description in the PR, and proper commit messages. That would save us both time. All I can do is guess what you actually try to accomplish/fix. Thank you.

PiotrProkop:
You guessed correctly,  I want to filter out all `diskless` pools. Balancer scheduler was created having NVMEoF in mind, so it is picking storage nodes in the same rack as NVME initiator and then tries to choose least use Node, Nic(NUMA node) and Storage Pool. In previous versions of Linstor, API filtered out `DfltDisklessStorPool`, so there was no `diskless` Storage Pools on storage nodes by default.

rck:
LGTM. I also talked to the LINSTOR devs, they confirmed that this change was intentional, so yes, there needs to be some kind of filtering if diskless pools should be excluded. I merged you change, thanks. I guess there will be a release pretty soon (guess next week).

PR #62, opened 2020-02-28
TLS mutual auth support

NeonSludge:
This PR implements TLS mutual auth support for the LINSTOR controller API endpoint.
You can now pass three additional environment variables to the `linstor-csi-plugin` containers:

1. `LS_USER_CERTIFICATE`: a client certificate that is trusted by the LINSTOR controller.
2. `LS_USER_KEY`: corresponding private key.
3. `LS_ROOT_CA`: certificate authority certificate (or chain) that the client will use to verify server certificates.

These variables might receive their values from a Kubernetes secret or be defined directly in the YAML spec.
All three variables must be set in order to activate TLS mutual auth. All of them must contain PEM-encoded data. If all three are set and some error occurs while parsing them, then `linstor-csi` fails with a fatal error.

### DIFF
--- a/cmd/linstor-csi/linstor-csi.go
+++ b/cmd/linstor-csi/linstor-csi.go
@@ -78,7 +78,7 @@ func main() {
 		caPool := x509.NewCertPool()
 		ok := caPool.AppendCertsFromPEM([]byte(caPEM))
 		if !ok {
-			log.Fatal("failed to get a valid certificate from LS_CA_PEM")
+			log.Fatal("failed to get a valid certificate from LS_ROOT_CA")
 		}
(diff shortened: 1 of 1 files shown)


rck:
Before I review the code, please bring it into a shape where it can be applied:
- `git reset --hard` your branch to before the Dockerfile changes. That was already fixed on master
- `git fetch` our remote
- `git rebase` your branch on our master
- `git push -f` your branch that now has removed the unnecessary Dockerfile chnage (and the also unnecessary revert (why didn't you squash these in the first place?)), and is rebased on master and should be in a state where it even can be cleanly applied.

I will review the code ASAP.
Thanks for you work, rck

NeonSludge:
> 
> 
> Before I review the code, please bring it into a shape where it can be applied:


Done, sorry about that :)

rck:
LGTM, pushed to master. I also squashed your two commits, hope you don't get payed per upstream commit :D

Thanks for your contribution, rck

PR #66, opened 2020-04-15
support for resourcegroups

rck:
add support for LINSTOR resource-groups

All options that can be set on a k8s storage class that are feasible for
a LINSTOR RG, are now set on said RG. All over this cleans up the code
nicely as we don't have to copy certain options over and over again.

It also allows setting DRBD properties like:
parameters:
  placementCount: "2"
  storagePool: "lvm-thin"
  resourceGroup: basic-tp
  DrbdOptions/Net/allow-two-primaries: "yes"

These options are exactly as defined by the LINSTOR REST API.

If the last PV of a RG is deleted, the RG is removed. Which is fine, it
will be recreated if necessary later on a PVC create.

'resourceGroup' is now a mandatory option in a SC. Usually this should
be unique and be set to the name of the SC. This is unfortunately
necessary, because the CSI driver does not passed in the name of the SC,
only the SC parameters.

Editing the SC (e.g., to remove/update DRBD options) is not possible, as
k8s does not allow editing of SC parameters. Choose them wisely :).

This also updates golinstor to the latest version, changing some details
in how logging worked.


pkg/client/linstor.go
```diff
@@ -315,7 +308,16 @@ func (s *Linstor) Delete(ctx context.Context, vol *volume.Info) error {
...
 
 	// No snapshots, remove the resource.
-	return nil404(s.client.ResourceDefinitions.Delete(ctx, vol.ID))
+
+	// try to rm the RG, but don't consider it a failure if we can not (maybe other resources,...)
+	rd, rdErr := s.client.ResourceDefinitions.Get(ctx, vol.ID)
+	retErr := nil404(s.client.ResourceDefinitions.Delete(ctx, vol.ID))
+
+	if rdErr == nil && rd.ResourceGroupName != "" {
+		_ = s.client.ResourceGroups.Delete(ctx, rd.ResourceGroupName)
```
JoelColledge, line 317: Do we really want to suppress all errors?

Also, won't linstor generate an error report whenever this fails due to other resources being present?

rck: I thought it saves a query/check and in the worst case it is a RG hanging around. But generating an error report is a very valid point. I will update this part.

JoelColledge:
LGTM

PR #70, opened 2020-05-03
Implement volume resizing

kvaps:
Based on https://github.com/piraeusdatastore/linstor-csi/pull/44#pullrequestreview-332592015

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- support `VolumeExpand` capabilities
+
+### Changed
+- Updated CSI spec to v1.2.0
 
 ## [0.8.3] - 2020-06-08
 ### Fixed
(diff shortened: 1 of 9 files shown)


WanzenBug:
Hey! I re-reviewed the code and tested it. Looks good from my side :+1: 
I took the liberty to rebase and squash everything into a single commit
@rck LGTM

rck:
All in all this LGTM. I did not follow the whole call path, but @WanzenBug, I'd assume it needs a "noop fs resizer" if it is in block mode, doesn't it?

WanzenBug:
@rck thanks for mentioning the block mode issue!

The original code did not handle this case. 

The problem is that the resizer module tries to be smart and detect the filesystem on the device. Now a newly created volume in block mode won't have any detectable FS, and the resizer will skip it. 

However, a user could run something like `mkfs.vfat` on the device, and the resizer would continuously try to resize, but return an error because `vfat` is not supported.

So we need a way to detect if the PVC is created in blockmode. In the CSI spec v1.1.0, we do not get access to the volume capability from the `*ExpandVolume()` calls. This was changed in v1.2.0 for exactly this use case.

Then we can tell the CSI infrastructure that we don't need additional resizing steps if the volume was in block mode: [relevant code section](https://github.com/piraeusdatastore/linstor-csi/pull/70/files#diff-756cc8d3ac1bea225cc27f4638970191R982)

So in summary:
* Update to CSI Spec v1.2.0
* On expansion, check if volume is in block mode (via volume capability, added in CSI v1.2.0):
  * if in block mode, just resize the volume in LINSTOR
  * if in FS mode, resize in LINSTOR and let CSI call `NodeExpand`, which resizes the file system

rck:
Thank you all!

PR #72, opened 2020-05-03
Drop secrets permission

kvaps:
This PR drops permissions for read secrets in all cluster for linstor-csi-provisioner and linstor-csi-snapshoter.
This is not required for LINSTOR.


kvaps:
Should I also update old manifests for `1.14`, `1.15`, `1.16` and `dev`?

rck:
Quite frankly I don't know how these got there in the first place (some copy/paste from somewhere I assume?), but I tend to agree that these are not necessary. 

I don't think it is that critical and we keep the old yamls as they are.

Thanks, rck

PR #73, opened 2020-05-14
allow empty storage pools

rck:
For LINSTOR it is perfectly fine that a RG does not contain a dedicated
storage pool. LINSTOR then chooses one.

In this case the Go string for the SP if not set in the SC had it's
default value (i.e., ""). This was then set on the RG, which then is not
valid in later on in LINSTOR.


JoelColledge:
Looks good to me.

Issue #68, opened 2020-04-20
The source volume and his snapshot are deleted when I try to create a new PVC from a VolumeSnapshot


SinHouse:
Hi, when I try to create a new PVC from a Volume Snapshot the source linstor volume and his snapshot are deleted and a new one linstor volume is created.

I follow [ kubernetes-1-17-feature-cis-volume-snapshot-beta](https://kubernetes.io/blog/2019/12/09/kubernetes-1-17-feature-cis-volume-snapshot-beta/) and the [CSI Snapshotter](https://github.com/kubernetes-csi/external-snapshotter) repo too.

I added the new CRDs from CSI Snapshotter repo by:
`kubectl create -f config/crd`
and the Snapshot Controller by:
`kubectl create -f deploy/kubernetes/snapshot-controller`

My config:
kubernetes v1.17.4
Kube-linstor: v1.6.0

Storage class:
```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: linstor
provisioner: linstor.csi.linbit.com
parameters:
  autoPlace: "3"
  storagePool: "linstor-pool"
```
Volume Snapshot Class:
```yaml
apiVersion: snapshot.storage.k8s.io/v1beta1
kind: VolumeSnapshotClass
metadata:
  name: linstor-snapshot
driver: linstor.csi.linbit.com
deletionPolicy: Delete
```
PVC:
```yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: test-volume
spec:
  storageClassName: linstor
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 500Mi
```
Volume Snapshot:
```yaml
apiVersion: snapshot.storage.k8s.io/v1beta1
kind: VolumeSnapshot
metadata:
  name: test-snapshot
spec:
  volumeSnapshotClassName: linstor-snapshot
  source:
    persistentVolumeClaimName: test-volume
```

From Linstor CLI:
```console
LINSTOR:~$ vd list
+---------------------------------------------------------------------------------------------+
| ResourceName                             | VolumeNr | VolumeMinor | Size    | Gross | State |
|=============================================================================================|
| pvc-d9037b05-7f9e-4e10-b3fb-f6ddc7f861d3 | 0        | 1000        | 500 MiB |       | ok    |
+---------------------------------------------------------------------------------------------+
LINSTOR:~$  s list
[...]

PR #75, opened 2020-06-03
manually create resource/volume definitions

WanzenBug:
Snapshots could not be restored since using ResourceGroups (RG). This was
because using the /spawn endpoint on RGs also creates a
VolumeDefinition (VD). On restoring the VD from the snapshot, linstor will
refuse, based on the fact that a VD with matching number already exists.

The fix is to manually create VDs instead of relying on /spawn.
The new volume creation process is:
* Create RG from volume params (== storage class params)
  There is only one RG for a storage class (as long as the RG name is set)
* Create ResourceDefinition for the new volume (and associate with RG)
* Create VolumeDefinition associated to ResourceDefinition

To create a new volume from a snapshot:
* Create RG from volume params (if not already present)
  There is only one RG for a storage class (as long as the RG name is set)
* Create ResourceDefinition for the new volume (and associate with RG)
* Restore VD from snapshot into ResourceDefinition
* Restore volumes based on VD and snapshot


WanzenBug:
Issue #68 is mostly fixed by just this line: https://github.com/piraeusdatastore/linstor-csi/pull/75/files#diff-756cc8d3ac1bea225cc27f4638970191R1072

However the above error only happens without all the "resource-group" business, so v0.7.4 or below

rck:
Thanks, that was a very cool contribution!

PR #76, opened 2020-06-08
ensure annotations on RDs are up-to-date

WanzenBug:
CSI Node needs annotations on resource definitions to determine
which volume to mount. On snapshot restore, LINSTOR will overwrite
all props of a RD with the original value. This lead to CSI Node mounting
the "original" volume instead of the one restored from a snapshot.

The fix is to move the creation/update of these annotations to the end
of the snapshot-restore method.


rck:
Thanks!

PR #83, opened 2020-09-01
Reduce created topology keys for volumes

WanzenBug:
the topology feature is essential when working with volumes that are not
remote attachable (for example, without a drbd layer). A node publishes
a set of labels that can later be referenced by volumes to bind to a
specific node set.

A created volume can be accessible via 1-2 "topologies":
* node names: this is used to directly relate the backing disks created
  in LINSTOR to a set of nodes. This works for all LINSTOR managed resources.
* storage pool names: this is used in case remote access to a volume is
  possible. This depends on the wheter the storage class supports diskless
  volumes.

Using storage pool names greatly decreases the size of the generated PV spec,
since only a few nodes are directly referenced (those with backing disk
resources). In the old appoach, only using node names meant that any "diskless"
supporting volume would:
* have a list of _all_ nodes in the cluster in its spec
* would not be schedulable on nodes that joined the cluster after the volume
  was created (the list would not be updated at that point).

This is a different approach then #56 in the face of #54 
Also see the discussion in the operator project: https://github.com/piraeusdatastore/piraeus-operator/issues/68


WanzenBug:
For example, a default PV created with this change would have this spec:
```yaml
    nodeAffinity:
      required:
        nodeSelectorTerms:
        - matchExpressions:
          - key: linbit.com/hostname
            operator: In
            values:
            - centos-7-k8s-100.test
        - matchExpressions:
          - key: linbit.com/sp-DfltDisklessStorPool
            operator: In
            values:
            - "true"
```

A volume with no support for "diskless" attaching would look like this:
```yaml
    nodeAffinity:                
      required:
        nodeSelectorTerms:
        - matchExpressions:
          - key: linbit.com/hostname
            operator: In
            values:
            - centos-7-k8s-100.test
```

pkg/topology/topology.go
```diff
@@ -39,6 +41,24 @@ const (
...
+	// The value assigned to the storage pool label, given that the node has access to the storage pool.
+	LinstorStoragePoolValue = "true"
+)
+
+// Converts the storage pool name into a CSITopology compatible label.
+//
+// For example: "abcd-efg" will be converted to "linbit.com/sp-abcd-efg".
+// There is an upper limit on the length of these keys, so the storage pool name may be truncated to fit the maximum
+// label length.
+func ToStoragePoolLabel(storagePoolId string) string {
```
rck, line 62: what is this limit and where is it defined? Can we use that value here? And I'd prefer to have the overall length then be calculated based on the length of this and `LinstorStoragePoolKeyPrefix`. Or glue them together and do a `[:max]`.

Is it possible that we generate collisions? `maxCSI < maxLINSTORSPname + len(LinstorStoragePoolKeyPrefix)`.

WanzenBug: Yeah, its a bit of a mess unfortunately. The limit is described in the CSI Spec (Look for `message Topology {`): https://github.com/container-storage-interface/spec/blob/v1.0.0/spec.md

Its defined as a "prefix" which is everything before the `/`, max 63 characters, and the actual key, which again is only allowed to contain 63 characters. Also this: 
> `Keys MUST be case-insensitive`

So I'm not sure how to proceed. We could hash the LINSTOR name, which should keep us under the 63 character limit without fear of collisions.

rck: the LINSTOR name is max 48 chars, so we should always be on the safe side. just document it here. maybe do a check+panic for the impossible case?

WanzenBug: Panic should be fine in this case.

pkg/topology/topology.go
```diff
@@ -39,6 +41,27 @@ const (
...
+
+	// The value assigned to the storage pool label, given that the node has access to the storage pool.
+	LinstorStoragePoolValue = "true"
+)
+
+// Converts the storage pool name into a CSI Topology compatible label.
+//
+// There is an upper limit on the length of these keys (63 chars for prefix + 63 chars for the key) as per CSI Spec.
+// LINSTOR enforces a stricter limit of 48 characters for storage pools, so this should not be an issue
+func ToStoragePoolLabel(storagePoolName string) string {
```
rck, line 61: sorry for being picky on that, but I'd like to have the check on the final generated string. something like:
```go
label := fmt.Sprintf("%s/sp-%s", LinstorStoragePoolKeyPrefix, storagePoolName)
if len(label) > 63+1+63 { // should be impossible
        panic(....)
}```

WanzenBug: Thinking about this again, this approach is essentially the same as not checking at all. In this case our pod panics, without the check the registrar plugin stops with a helpful error message: 
> `Invalid value: "linbit.com/sp-....": name part must be no more than 63 characters`
So I'd be in favour of not checking at all.

rck: yes, fine with me, just keep what we already documented, that it is impossible, and that it would fail anyways.

PR #84, opened 2020-09-09
Update dev deployment to be up-to-date with 1.17

WanzenBug:
The last dev deployment example was not updated to support newer
kubernetes versions. Newer kubernetes distributions removed support
for some of the API versions which were still in use. This commit
updates the dev deployment to be in line with the 1.17 deployment.

Also adds a clarification that this project only deploys the CSI
components and links to the operator for more information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,11 @@ run `make staticrelease`.
 
 # Deployment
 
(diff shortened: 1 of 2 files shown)


rck (approved):
LGTM

PR #85, opened 2020-09-14
Allow all available Linstor layers in storage class

WanzenBug:
Add support for the CACHE, OPENFLEX and WRITECACHE layers.
Also replace deprecated "DISKLESS" flag with more specific
version based on the available layers.


rck (approved):
LGTM, thanks for also writing tests!

PR #87, opened 2020-10-06
Add support for ROX volumes

WanzenBug:
DRBD and by extension LINSTOR support ROX configurations, as
long as the device is mounted with the "ro" option. This commit enables
the CSI driver to provision such ROX volumes. For this to work we need to:
* tell CSI we can handle ROX configurations
* validate ROX volumes are used with the "readOnly" option set


WanzenBug:
@rck Ready for review. Block volume support for ROX is not included for now.

rck (approved):
LGTM

PR #88, opened 2020-10-09
add support for ROX block volumes

WanzenBug:
Block volumes are special, extra care must be taken such that they
are not accidentally opened RW in ROX mode. In our setup, block volumes
are exposed to Kubernetes as simple bind mounts (with "ro").

However, it is still possible to open DRBD devices as O_RDWR through the
bind mount. In a later stage, Kubernetes sets up a loop device using
losetup for our block volume. By default, losetup opens the device
as O_RDWR if it can. This means we need a way to disallow O_RDWR on
our exported devices.

The solution implemented here is to set up _another_ loop device, using
losetup's "--read-only" option to ensure the backing device is only opened
using O_RDONLY.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
-- Support for ROX filesystem volumes ([#87])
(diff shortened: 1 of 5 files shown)


rck (approved):
Code LGTM. I'm a bit worried that it adds "complexity" when nobody asked for that feature (and quite frankly I don't see that coming for block devices), but who knows.

ACK, you hit the button :)

Issue #91, opened 2020-10-29
fsGroup does not apply


mxsavchenko:
**Linstor csi:** 0.9.1
**Kubernetes:** v1.18.3.
**Pool:** ZFS.
**DRBD:** DRBDADM_API_VERSION=2 DRBD_KERNEL_VERSION=9.0.25 DRBDADM_VERSION=9.15.0

I use securityContext to run not as root user and i have no write access to the mounted volume.
My test Pod:
```
kind: StorageClass
apiVersion: storage.k8s.io/v1
metadata:
  name: linstor-hdd-r2
parameters:
  autoPlace: "2"
  storagePool: hdd
provisioner: linstor.csi.linbit.com

---
kind: PersistentVolumeClaim
apiVersion: v1
metadata:
  name: task-pv-claim
spec:
  storageClassName: linstor-hdd-r2
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 8Gi

---
kind: Pod
metadata:
  name: task-pv-pod-test2
spec:
  securityContext:
    runAsUser: 1000
    fsGroup: 1000
  volumes:
    - name: task-pv-storage
      persistentVolumeClaim:
        claimName: task-pv-claim
  containers:
    - name: task-pv-container
      image: busybox
      command:
        - sh
        - -c
        - 'while true; do date > /mnt/index.html; sleep $(($RANDOM % 5 + 5)); done'
      securityContext:
        runAsUser: 1000
      volumeMounts:
        - mountPath: "/mnt"
          name: task-pv-storage
```
result:
```
k exec -it task-pv-pod-test2 -- ls -al | grep mnt
drwxr-xr-x    3 root     root          4096 Oct 28 08:55 mnt

---
stat  /mnt
  File: /mnt
  Size: 4096            Blocks: 8          IO Block: 4096   directory
Device: 409305h/4231941d        Inode: 2           Links: 3
Access: (0755/drwxr-xr-x)  Uid: (    0/    root)   Gid: (    0/    root)

---
k logs task-pv-pod-test2
sh: can't create /mnt/index.html: Permission denied
sh: can't create /mnt/index.html: Permission denied
sh: can't create /mnt/index.html: Permission denied
```

kubelet logs:
```
docker logs kubelet 2>&1 | grep pvc-182197ac-2d36-47ac-8214-8245db8fea2b
[...]

PR #92, opened 2020-10-29
use explicit fstype in storageclass example

WanzenBug:
When using newer CSI components, no default fstype is set for PVs.
This leads to kubelet skipping the fsGroup step since it expects
the volume to be in block mode.

See https://github.com/piraeusdatastore/linstor-csi/issues/91


rck:
LGTM. we might have some examples in `linbit-documentation` that we should update as well.

PR #93, opened 2020-11-18
remove losetup dependency for ROX support

WanzenBug:
In d2f23d6 we introduced ROX volume support, which works around DRBD's
autopromote feature by introducing a readonly loop-device so all access
is done readonly.

The code to handle detaching needed to clean up the loop device, which
relied on a specifc implementation of "losetup --list", which is not
compatible with older version (such as the one found in CentOS 7).
This meant that _all_ PVC failed to detach.

This commit switches to using "blockdev --setro" to ensure readonly access.
That in turn requires DRBD v9.0.26, which is not released yet. This should
not be a problem, as the feature was probably never used anyways.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- ROX block volumes support requires DRBD v9.0.26 ([#93])
+
+[#93]: https://github.com/piraeusdatastore/linstor-csi/pull/93
+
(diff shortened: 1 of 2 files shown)


WanzenBug:
@rck I guess you were right about this feature :/

rck:
LGTM. Thanks for fixing it

WanzenBug:
That has nothing to do with LINSTOR CSI (we don't use `losetup` anywhere). For some reason kubelet _always_ creates a loop device for `Block` volumes, even if the exported bind target is already a block device

kvaps:
Got it! Thanks

PR #94, opened 2020-11-24
Update CSI sanity tests

WanzenBug:
* Update csi-test package to 3.1.1
* Use 'sanity.' prefix for test binary options
* Add option to generate XML report from test run.
* Run `go test ./...` in CI

PR #95, opened 2020-11-25
deal with non-existent node publish paths

WanzenBug:
* Wrong paths return the expected NotFound error in NodeGetVolumeStats
* Remove created mount point file/dir in NodeUnpublishVolume
* Make NodeUnpublishVolume indempotent

### DIFF
--- a/pkg/client/linstor.go
+++ b/pkg/client/linstor.go
@@ -24,6 +24,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"os"
 	"regexp"
 	"strings"
 
@@ -1124,8 +1125,19 @@ func (s *Linstor) setDevReadWrite(srcPath string) error {
 }
 
 // IsNotMountPoint determines if a directory is a mountpoint.
+//
+// Non-existent paths return (true, nil).
 func (s *Linstor) IsNotMountPoint(target string) (bool, error) {
-	return s.mounter.IsNotMountPoint(target)
+	notMounted, err := s.mounter.IsNotMountPoint(target)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return true, nil
+		}
+
+		return false, err
+	}
+
+	return notMounted, nil
 }
 
 // Unmount unmounts the target. Operates locally on the machines where it is called.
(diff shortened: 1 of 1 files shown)


WanzenBug:
Wait for #94 to be merged

pkg/client/linstor.go
```diff
@@ -1124,8 +1125,15 @@ func (s *Linstor) setDevReadWrite(srcPath string) error {
...
+//
+// Non-existent paths return (true, nil).
 func (s *Linstor) IsNotMountPoint(target string) (bool, error) {
-	return s.mounter.IsNotMountPoint(target)
+	notMounted, err := s.mounter.IsNotMountPoint(target)
+	if err != nil && !os.IsNotExist(err) {
+		return false, err
+	}
+
+	return notMounted, nil
```
rck, line 1136: So there is the case where it was an `err`, and where it was an `os.IsNotExist()`, then you return what `IsNotMountPoint` gave you for `notMounted`. Even if that works in this particular implementation, the common rule/"contract" IMO is: if it es an `error`, you can not rely on the other return values what so ever, it is allowed to be anything.

Please rephrase that.

pkg/client/linstor.go
```diff
@@ -1134,18 +1142,23 @@ func (s *Linstor) Unmount(target string) error {
...
+		// Still needs to be unmounted
+		err := s.mounter.Unmount(target)
+		if err != nil {
+			return fmt.Errorf("unable to unmount target '%s': %w", target, err)
+		}
 	}
 
-	err = s.mounter.Unmount(target)
+	// We create a file/directory in Mount() which must be removed again
+	out, err := s.mounter.Run("rm", "-fd", target)
```
rck, line 1159: we are all GNU/Linux specific anyways, here and in LINSTOR and what not, but I'd prefer a "Go solution" if it is possible. Guess that should not be too cumbersome in this case.

rck (changes requested):

pkg/client/linstor.go
WanzenBug: I wasn't sure if that would always work: there is some magic going on in the `mounter.Run()` method regarding namespaces. But that does not seem to be active in our containers, so a normal `os.Remove()` works :shrug:

pkg/client/linstor.go
WanzenBug: You are absolutely right on that, updated it just now.

rck:
thanks!

PR #101, opened 2020-12-11
Snapshot refactor

WanzenBug:
The changes in detail:

Instead of storing snapshot information in resource definitions, use native
LINSTOR snapshots. This makes it easier to retry a timed out snapshot attempt.
A basic test was added using "old" format LINSTOR responses to ensure no
information is lost in transition.

To ensure stable snapshot ID between tries, the suggested name is either
used directly or a generated UUIDv5 based on the name is used. Since the CSI
snapshotter image will always suggest a compatible name, this is mostly done
for complying with the CSI spec.

In volume-from-volume scenario, re-use the snap-create and vol-from-snap logic
instead of implementing it in the LINSTOR client. This is done by first creating
a stable snapshot name for the new PVC, restoring using the existing
vol-from-snap logic and deleting the snapshot again.

The snapshots itself are only considered ready if LINSTOR reports the SUCCESS
flag. This prevents a lot of churn with snapshots that are not immediatly ready
being deleted over and over again.

Pagination is handled directly by the client, as we use the /view/snapshots
API directly.

### DIFF
--- a/pkg/client/linstor.go
+++ b/pkg/client/linstor.go
@@ -24,15 +24,23 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math"
 	"os"
 	"regexp"
 	"strings"
 
+	lapiconsts "github.com/LINBIT/golinstor"
 	lapi "github.com/LINBIT/golinstor/client"
 	"github.com/container-storage-interface/spec/lib/go/csi"
-	ptypes "github.com/golang/protobuf/ptypes"
+	"github.com/golang/protobuf/ptypes/timestamp"
 	"github.com/haySwim/data"
 	"github.com/pborman/uuid"
+	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
+	"k8s.io/kubernetes/pkg/util/mount"
+	"k8s.io/kubernetes/pkg/util/resizefs"
+	"k8s.io/kubernetes/pkg/util/slice"
+
 	"github.com/piraeusdatastore/linstor-csi/pkg/linstor"
 	lc "github.com/piraeusdatastore/linstor-csi/pkg/linstor/highlevelclient"
 	"github.com/piraeusdatastore/linstor-csi/pkg/linstor/util"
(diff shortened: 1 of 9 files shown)


WanzenBug:
@rck do you want to start the review? I'm still waiting for some internal tests to finish up, the last run already shows more tests passing then before this PR, so I expect no further changes to this PR

pkg/client/linstor.go
```diff
@@ -879,96 +843,133 @@ func (s *Linstor) ListVolumes(ctx context.Context) ([]*volume.Info, error) {
...
+		return nil, fmt.Errorf("missing volume definitions")
+	}
+
+	if len(lsnap.Snapshots) == 0 {
+		return nil, fmt.Errorf("missing snapshots")
 	}
-	close(allSnaps)
 
-	<-done
+	snapSize := lsnap.VolumeDefinitions[0].SizeKib * 1024
```
rck, line 962: maybe `snapSizeBytes`

pkg/driver/driver.go
```diff
@@ -864,69 +901,66 @@ func (d Driver) DeleteSnapshot(ctx context.Context, req *csi.DeleteSnapshotReque
 
 // ListSnapshots https://github.com/container-storage-interface/spec/blob/v1.2.0/spec.md#listsnapshots
 func (d Driver) ListSnapshots(ctx context.Context, req *csi.ListSnapshotsRequest) (*csi.ListSnapshotsResponse, error) {
-	var snapshots = make([]*volume.SnapInfo, 0)
+	limit := int(req.GetMaxEntries())
+	start, err := parseStartingToken(req.GetStartingToken())
```
rck, line 905: "parse" is a bit common. Maybe something that transport the "convert to int"? or `var start int`. It is just not clear what that `start` could be without jumping around.

rck (changes requested):
The overall approach LGTM. ZFS snapshots quirks got discussed elsewhere and the README was updated. The rest is just some nitpicking, the `break` should be added.

pkg/driver/driver.go
WanzenBug: renamed to `parseAsInt`

pkg/client/linstor.go
```diff
@@ -879,96 +843,133 @@ func (s *Linstor) ListVolumes(ctx context.Context) ([]*volume.Info, error) {
...
+
+	opts := &lapi.ListOpts{
+		Offset: start,
+		// if set to 0 the offset parameter is ignored, so we use MaxInt as substitute for "unlimited"
+		Limit: math.MaxInt32,
 	}
 
-	return s.doGetSnapByID(vols, id), nil
-}
+	if limit > 0 {
```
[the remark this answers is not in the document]
WanzenBug: reworked, by setting the `limit` argument to `MaxInt32` before filling `opts`

rck (approved):
LGTM

PR #106, opened 2021-01-04
add "nouuid" option to XFS mounts

WanzenBug:
Restored snapshots inherit the XFS UUID of the original source. If mounted
on the same node as the original without the "nouuid" option, XFS will
complain about a duplicate UUID and refuse to mount.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- Add `nouuid` to default XFS mount options. This enables mounting restored snapshots on the same node as the original.
+
 ## [0.11.0] - 2020-12-21
 
 ### Changed
--- a/pkg/driver/driver.go
+++ b/pkg/driver/driver.go
@@ -300,6 +300,12 @@ func (d Driver) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolum
 		}
 	}
 
+	if fsType == "xfs" {
+		// Restored snapshots inherit the XFS UUID of the original source. If mounted on the same node as the original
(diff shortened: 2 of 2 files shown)


rck:
Hm, or should we try to generate a new one? maybe should even LINSTOR do that for us?

Should we have this one as "quick fix" as it is trivial and the other one as LINSTOR feature request?

WanzenBug:
Since LINSTOR is only concerned about the block volumes, I don't see any clear path to add this as a new feature.

I also considered the `xfs_admin -U generate` route, but that would add a new complication to the read-only mode: There would need to be a window where only one node executes the `xfs_admin` command, while the others wait for the change to propagate.

This second issue is a result of the way we lazily initialize the file system. If we could find a way to manage the file system already in the provisioning stage, we could also run `xfs_admin` there.

rck:
> Since LINSTOR is only concerned about the block volumes, I don't see any clear path to add this as a new feature.

actually it isn't. It has a feature to create file systems. So conceptionally it could also extend that to a "if the parent had the 'create file system option' set, then also make sure the children have a new UUID".

> There would need to be a window where only one node executes the xfs_admin command, while the others wait for the change to propagate.

True, LINSTOR would be that central lock

> This second issue is a result of the way we lazily initialize the file system.

That could be done by LINSTOR I guess. The "guess" is the important part here, I'd really have to look into when the FS is created how, especially with snapshots.

Don't know if you knew that LINSTOR can create FSs with options and if that changes anything. The fix as is is fine with me, feel free to merge. Just (re)consider if moving FS creation/UUID generation to LINSTOR is possible.

WanzenBug:
> actually it isn't. It has a feature to create file systems
TIL

I'll have to look into this LINSTOR feature. For now I'll merge this while thinking about how we could benefit from using LINSTOR in this case.

PR #107, opened 2021-01-13
Make snapshots independent from their source PV

WanzenBug:
This commit enables deletion of PV while taken snapshots are kept around.
To implement this a couple of changes needed to be made:

Snapshots in LINSTOR are attached to a ResourceDefinition (RD), not to the
Resources directly. This means we can remove a PV by deleting the Resources
in LINSTOR without deleting the RD.

In turn, we cannot rely on the "volume.Info" struct to exist when
restoring from a snapshot. While it was queried, it actually did not serve
any purpose, the query is just removed.

Not deleting the RD on PV deletion requires another set of changes:
* Once the last RD "user" (Resource or Snapshot) is deleted, we need to
  delete the RD (and maybe also the Resource Group)
* We use aux props on the RD to persist important metadata. Specifically,
  it was also used to identify RDs and their matching PV. This metadata
  is now stripped on volume deletion.

### DIFF
--- a/pkg/client/linstor.go
+++ b/pkg/client/linstor.go
@@ -311,10 +311,10 @@ func (s *Linstor) Create(ctx context.Context, vol *volume.Info, req *csi.CreateV
 		return err
 	}
 
-	// saveVolume() makes the volume eligible for further use, i.e. this method will not be called again once saveVolume
+	// saveVolumeMetadata() makes the volume eligible for further use, i.e. this method will not be called again once saveVolumeMetadata
 	// succeeded.
 	logger.Debug("update volume information with resource definition information")
-	if err := s.saveVolume(ctx, vol); err != nil {
+	if err := s.saveVolumeMetadata(ctx, vol); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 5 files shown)


WanzenBug:
I also had to update the way we handle resource groups, otherwise on every "snapshot restore" the old resources would be re-created in LINSTOR.

pkg/volume/volume.go
```diff
@@ -237,49 +236,83 @@ func NewParameters(params map[string]string) (Parameters, error) {
...
+
+	if !reflect.DeepEqual(rg.SelectFilter.ReplicasOnSame, params.ReplicasOnSame) {
+		changed = true
+		rgModify.SelectFilter.ReplicasOnSame = params.ReplicasOnSame
+	}
 
-	if params.StoragePool != "" { // otherwise we set the storagepool to "", which does not make LINSTOR happy
-		rgModify.SelectFilter.StoragePool = params.StoragePool
-		rgModify.OverrideProps[lc.KeyStorPoolName] = params.StoragePool
+	if !reflect.DeepEqual(rg.SelectFilter.ReplicasOnDifferent, params.ReplicasOnDifferent) {
```
rck, line 270: besides my some general dislike of `reflect`, is this really necessary? these are just string slices IIRC. also it could result in `updated = true` if only the order (but not the semantic) changes, right? all in all I think I would use simple helper function checking for `len()` and if every x is in y.

pkg/client/linstor.go
```diff
@@ -1341,3 +1297,71 @@ func (s *Linstor) GetNodeTopologies(ctx context.Context, nodename string) (*csi.
...
+
+	rDef, err := s.client.ResourceDefinitions.Get(ctx, rdName)
+	if err != nil {
+		// if no RD was found, returns nil
+		return nil404(err)
+	}
+
+	log.Debug("checking for undeleted resources")
+
+	resources, err := s.client.Resources.GetAll(ctx, rdName)
```
rck, line 1316: if the comment in the err case is correct, then you don't need the part above (`ResourceDefinitions.Get()`). You need the `rDef` later at the end. You could move the `rDef` part down. IMO it would make it a bit easier to read.

rck (commented):
only some minor feedback you might want to take into consideration. I like the overall approach, good work

rck (commented):
only some minor things you might want to take into consideration. I like the approach, good work

pkg/volume/volume.go
WanzenBug: :+1: added a simple helper function. While order is not important for `replicasOnSame` and `replicasOnDifferent`, it can be important for the layerstack. In any case, these values shouldn't even change, so I think its fine to assume they are always in the same order

pkg/client/linstor.go
WanzenBug: Done. Almost moved that part to after it was already deleted :sweat_smile:

PR #110, opened 2021-01-20
only delete resources on Detach if created in Attach

WanzenBug:
We want to delete resources that were only created to facilitate an Attach()
request. The way this was determined was by checking if the given resource is
diskless or not. This presents a problem, as there are a number of ways
diskless resources get created:
* Linstor creates a TieBreaker resource
* The user specified that a Diskless resource should be placed on that node
* Attach() was called for a node were the resource was not present

From the point of view of Detach(), all these above cases look the same, and
so the resources were deleted even if they did not originate from the Attach.

This commit adds a new property to resources generated by Attach. This way
Detach can later check for the presence of this property, and only delete
the resource if:
* The resource has the "Attach" property
* The resource is still diskless (there could be cases in which LINSTOR
  migrated some replicas from a evicted node) Since we don't want to mess
  with the replication factor, diskfull resources are off-limits.

### DIFF
--- a/pkg/client/linstor.go
+++ b/pkg/client/linstor.go
@@ -414,35 +414,36 @@ func (s *Linstor) Attach(ctx context.Context, vol *volume.Info, node string) err
 	if err != nil {
 		return err
 	}
+
+	rc.Resource.Props[linstor.PropertyCreatedFor] = linstor.CreatedForTemporaryDisklessAttach
 	return s.client.Resources.Create(ctx, rc)
 }
 
 // Detach removes a volume from the node.
 func (s *Linstor) Detach(ctx context.Context, vol *volume.Info, node string) error {
+	log := s.log.WithFields(logrus.Fields{
+		"volume":     fmt.Sprintf("%+v", vol),
+		"targetNode": node,
+	})
+
 	res, err := s.client.Resources.Get(ctx, vol.ID, node)
-	if err == lapi.NotFoundError {
-		s.log.WithFields(logrus.Fields{
-			"volume":     fmt.Sprintf("%+v", vol),
-			"targetNode": node,
-		}).Info("volume does not even exist")
-		return nil
-	} else if err != nil {
-		return err
+	if err != nil {
+		// if the resource could not be found there is nothing to detach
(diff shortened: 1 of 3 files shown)


pkg/client/linstor.go
```diff
@@ -414,35 +414,36 @@ func (s *Linstor) Attach(ctx context.Context, vol *volume.Info, node string) err
...
+		// if the resource could not be found there is nothing to detach
+		return nil404(err)
 	}
 
-	s.log.WithFields(logrus.Fields{
-		"resource":   fmt.Sprintf("%+v", res),
-		"targetNode": node,
-	}).Info("detaching volume")
+	createdFor, ok := res.Props[linstor.PropertyCreatedFor]
+	if !ok || createdFor != linstor.CreatedForTemporaryDisklessAttach {
```
rck, line 436: hm, what about legacy compat? wouldn't that keep the ones that truly have been created for attach only around forever? do users really create intentional/manual diskless assignments in a CSI context? if not, then we could delete resources that don't have the property *and* are diskeless *and* are not tie breakers. Does LINSTOR expose that information if it was a created as tie breaker? maybe we should request such a "tag" from the LINSTOR folks?

All that said, I'm fine if you merge it like this, keeping some unnecessary diskless resources should not hurt that much, probably not worth the effort doing some legacy compat here

WanzenBug: Yeah, this bit is a compromise. From my tests, LINSTOR removes the `TIE_BREAKER` flag from a resource once it is InUse. So after the volume is attached, we have no way to differentiate a diskless resource created as a tie breaker and a diskless resource created by CSI.

While we could argue that is a bug/missing feature in LINSTOR, I'd much rather wait for a better API for this use case in LINSTOR. Basically I just want to tell LINSTOR: I need this resource on that node, and later:  I don't need this anymore. LINSTOR can then figure out what to do based on the autoplace configuration.

rck: True, I'm always in shifting responsibility towards LINSTOR :). jokes aside, you are obviously right, that should be actually handled by LINSTOR in some improved way

PR #111, opened 2021-01-20
fix deletion order of diskless resources

WanzenBug:
Diskless resources need to be deleted before all diskfull resources are gone.
Otherwise the last diskfull resources can't ever be deleted. The easiest
way to ensure that diskfull resources are deleted last is by sorting the
resources beforehand.

Bug was introduced in #107.


rck:
good catch!

PR #115, opened 2021-04-22
Remove failed attempts at creating a snapshot

WanzenBug:
In some cases snapshots are created but fail to become "successful",
for example when one of the satellite goes offline at the wrong moment.
In such cases, we want to delete the failed attempt, as otherwise
the CSI snapshotter will forever wait for the snapshot to get ready.
This then produces a lot (read: gigabytes) of logs, as it tries polling
the snapshot state again multiple times per second.

This commit adds logic to recognize such failed snapshots and remove+retry
them. Note that currently there is no way to tell the snapshot controller that
a snapshot can't ever succeed, for example when a storage driver without
snapshot support is used.

### DIFF
--- a/pkg/client/linstor.go
+++ b/pkg/client/linstor.go
@@ -817,7 +817,7 @@ func (s *Linstor) deleteVolumeMetadata(ctx context.Context, vol *volume.Info) er
 	return nil
 }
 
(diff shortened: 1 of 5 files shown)


rck (approved):
LGTM

Issue #114, opened 2021-04-19
zfs ProvisioningFailed


aland-zhang:
linstor-csi: 0.12.1
drbd+zvol ,StorDriver/ZfscreateOptions Refer to this document configuration 'https://app.swaggerhub.com/apis-docs/Linstor/Linstor/1.5.0#/developers/resourceDefinitionModify'
Added the following StorageClass:

```
allowVolumeExpansion: true
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: drbd-zfs-2
parameters:
  StorDriver/ZfscreateOptions: "-o volblocksize=64k -o dedup=on"
  DrbdOptions/Net/allow-two-primaries: "no"
  allowRemoteVolumeAccess: "false"
  csi.storage.k8s.io/fstype: xfs
  disklessOnRemaining: "false"
  layerlist: drbd storage
  mountOpts: noatime,discard
  placementCount: "2"
  placementPolicy: 'AutoPlace'
  DrbdOptions/Net/protocol: "B"
  resourceGroup: drbd-zfs
  storagePool: pool_zfs
provisioner: linstor.csi.linbit.com
reclaimPolicy: Delete
volumeBindingMode: Immediate
```

Then after adding pvc, the following error is displayed:
`failed to provision volume with StorageClass "drbd-zfs-2": rpc error: code = Internal desc = CreateVolume failed for pvc-09db3844-5089-47bc-b1c2-077dea17a711: invalid parameter: stordriver/zfscreateoptions does not belong to paramKey values`

WanzenBug:
This is to be expected: currently only some of the properties can be set in the storage class. The relevant code is [here](https://github.com/piraeusdatastore/linstor-csi/blob/fd959147dba1f4cf6fe75fa1436ab327d8c348a6/pkg/volume/volume.go#L128)

As a workaround, you can set the option directly on the resource group: `linstor resource-group set-property StorDriver/ZfscreateOptions "-o volblocksize=64k -o dedup=on"`

aland-zhang:
Okay, but I also hope to add this configuration to StorageClass

PR #117, opened 2021-05-10
Allow all resource group properties in storage class

WanzenBug:
Allow setting all resource group level properties from storage classes.
The storage class parameters are exposed as the "params" of a volume request,
which internally get translated to a resource group.

To be as compatible as possible, every parameter prefix with
"property.linstor.csi.linbit.com" directly translates to a LINSTOR property.

Since this required reworking of resource groups, also make two further changes:
* Generate a unique resource group name based on the group properties
* Disallow modification of properties via CSI requests once set. There should
  never be 2 storage classes with the same resource group name anyways.


WanzenBug:
Would also fix #114

rck (commented):
CHANGELOG.md and the code contain trainling whitespace. You might want to fix that.

Code looks good an a first view

PR #119, opened 2021-05-17
toggle shared storage volume status on attach

WanzenBug:
Shared storage pools can lead to "inactive" replicas, which need to
be activated on a node before they can be used. First, this requires
existing volumes on the same pool to be deactivated, afterwards the
volume can be set active on the right volume.


rck:
guess you will try to satisfy the linter? just ping me when you are satisfied and then I will look at it in more detail

WanzenBug:
@rck ready. Since the upgrade of golinstor required changing some constants the linter felt it was necessary to complain about some "old" issues. Sorry for all the noise in the diff :/

pkg/volume/volume.go
```diff
@@ -335,33 +336,32 @@ func (params *Parameters) ToResourceGroupModify(rg *lapi.ResourceGroup) (lapi.Re
...
 			return lc.FlagDrbdDiskless, nil
-		}
-		if l == lapi.NVME {
-			return lc.FlagNvmeInitiator, nil
-		}
-		if l == lapi.OPENFLEX {
-			// Openflex volumes are connected via NVMe
+		case devicelayerkind.Nvme, devicelayerkind.Openflex:
 			return lc.FlagNvmeInitiator, nil
+		case devicelayerkind.Exos, devicelayerkind.Writecache, devicelayerkind.Cache, devicelayerkind.Storage, devicelayerkind.Luks:
```
rck, line 344: maybe just a default case? we do not really get a benefit if we enumerate all of them here, or do we? like when we add a new one we don't get some warning that it is not handled or anything, right?

rck (approved):
Besides the `case` thingly it looks reasonable to me. Feel free to fix that up or commit as is.

pkg/volume/volume.go
WanzenBug: Tell that to the linter ;-)

WanzenBug: Replaced by `default:` + setting `default-signifies-exhaustive: true` on the linter. Since we only check "new" issues, this linter would go out-of-date really quickly anyways the next time someone updates golinstor.

PR #121, opened 2021-06-09
Try to ensure quorum is achieved when attaching

WanzenBug:
In certain circumstances it is necessary to create a diskfull resource to make
it usable. The bug report that introduced this variable is a good example:
* Our cluster has 4 identical nodes called A, B, C, D.
* A resource was placed in a typical 2 + 1 Tiebreaker configuration, let's say
  the diskfull resources are on A and B, the Tiebreaker on C.
* A Pod attaches on node A.
* Node A goes down/becomes unreachable/etc. That leaves us with 1 diskfull
  resource on B + 1 diskless on C.
* The Pod is deleted and a replacement scheduled on node D (i.e. after deletion
  by the HA Controller).
* Now, this method will be called, since we need to Attach on node D, and there
  is currently no resource on that node.
* Using a diskless resource will not work, as the newly creates resource would
  not get quorum. This is because the new diskless will only see 1 available
  and one unavailable replica, so it can't decide which is in the majority.
* Using a diskfull resource will work and sync up to the remaining diskfull
  resource on B, as in that case the new resources itself already counts
  for the quorum majority.

For this change we need to detect these situations. In general, the promotion
score reported by LINSTOR is a good indicator of "unavailable" vs "available"
replicas, even if the satellite itself is reachable.

This commit also adds some tests for the above procedure and other edge cases
on calling Attach(). The LINSTOR API is mocked with responses taken from a real
LINSTOR instance exhibiting these edge cases.


rck:
LGTM on a first, quick look, but I wonder if we really need to handroll this? Can't we just let LINSTOR handle this? like re-calling auto-place or some "rebalance" they might have invented in the meantime? Probably we would need to tell LINSTOR that the node was lost for that, which brings me to the next question: what if the node comes back? we have over-redundancy now, right?

Maybe/most certainly it is worth it, but we are covering a failure mode the user decided to not cover, right? If you want to tolerate a node loss in such a  quorum case, you should have started with 1 more diskful replica in the first place... But I also get it that it might be nice to cover such cases without having an extra replica in all cases, I'm just not 100% convinced yet :)

WanzenBug:
> Can't we just let LINSTOR handle this?

Sure, I think eventually `Attach` will move to using `/make-available`. My own tests showed that this is still buggy right now, so I'm inclined to let it mature a bit.

> what if the node comes back? we have over-redundancy now, right?

Yes. However the extra replica will be deleted on `Detach`, so the replica will be deleted after use. Which reminds me that we should probably check that the original volume wasn't deleted in the meantime....

> but we are covering a failure mode the user decided to not cover, right?

Kinda? It works fine if you are always using 3 replicas. The issue is mainly experienced when using 2 replicas, which is also the most common configuration (and the one we use in most of our examples). I'd expect our example to "just work" :tm:

WanzenBug:
> However the extra replica will be deleted on Detach, so the replica will be deleted after use

Just checked, and that's only the case if the resource is diskless. So in the stated situation, the extra replica will _not_ be deleted. Which I guess is fine. Maybe someday we will get `linstor rd adjust` or `rebalance` or whatever and Linstor can figure out what replicas/resources need to be kept...

rck:
All in all it is the old question of how much knowledge we can expect. Still not entirely convinced that we should try to help here when the user "intentionally" decided against it. you get what you signed up for... but I also see that this is a common problem users might have and the patch looks simple enough, even if I would have preferred shifting that to LINSTOR, but that might happen eventually.

So in the end fine with me. code looked okay, so this is ready right?

WanzenBug:
Yeah, I'm still waiting for the internal e2e tests to pass. I'll ping you when it's done

WanzenBug:
e2e tests all passed. @rck ready

PR #122, opened 2021-06-14
Update to CSI spec 1.4.0

WanzenBug:
This commit only bumps the spec, ensures the driver conforms to the interfaces
and passes sanity tests, no functionality was changed or added.


pkg/driver/driver.go
```diff
@@ -1087,6 +1092,11 @@ func (d Driver) ControllerExpandVolume(ctx context.Context, req *csi.ControllerE
 	}, nil
 }
 
+// ControllerGetVolume https://github.com/container-storage-interface/spec/blob/v1.4.0/spec.md#controllergetvolume
+func (d Driver) ControllerGetVolume(ctx context.Context, request *csi.ControllerGetVolumeRequest) (*csi.ControllerGetVolumeResponse, error) {
+	panic("implement me")
```
rck, line 1097: that really is intentional?

WanzenBug: You are right, better to respond with a proper `Unimplemented` error

PR #123, opened 2021-06-18
Add bearer authentication for HTTP API

WanzenBug:
If the LINSTOR HTTP API is secured by a project like kube-rbac-proxy, which
secures an endpoint by delegating authentication to the kubernetes built-ins,
namely serviceaccounts and RBAC rules, we need a way to send our
authentication in the form of a bearer token.

On start-up, the token is read from a file specified on the command line. The
Header is added in a specialized http.RoundTripper implementation, since
golinstor currently does not support setting default headers.

This implementation even has the benefit of not logging the token on every
request.


rck (approved):
LGTM

PR #124, opened 2021-06-22
Make "FollowTopology" adhere strictly to CSI spec

WanzenBug:
The CSI spec requires very specific behaviour when using
AccessibilityRequirements/Topology.

A CreateVolumeRequest contains:
* A list of requisite topologies. The volume _has_ to be accessible
  from at least on of these. If the list is empty, any placement is allowed.
* A list of preferred topologies. If a list of requisites is also given,
  preferred is a subset of the requisite topologies. They should be considered
  first when placing volumes, but there is no requirement that a volume
  has to be accessible from a preferred topology.

This commit adds support for requisite and/or preferred topologies when using
the "FollowTopology" policy.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/piraeusdatastore/linstor-csi
 go 1.12
 
 require (
-	github.com/LINBIT/golinstor v0.34.3
+	github.com/LINBIT/golinstor v0.34.4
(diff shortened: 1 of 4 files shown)


rck:
what a great spec :D. I just needed some pen and paper, but yes, LGTM.

PR #125, opened 2021-07-13
Update go dependencies

WanzenBug:
Removed dependency on k8s.io/kubernetes and synced the other k8s dependencies
to the latest release. This required adding some dummy contexts and the
change of the following checks on volume attach:

* The "PathIsDevice" check. No longer available, can be replaced by a simple
  call to os.Stat().

* The "DeviceOpened" check. Removed, as it had the classic check-then-execute
  race conditition. Later mount calls do the same check in any case. The only
  time when the check could be useful is in case of mounting block volumes.
  Since block volumes are bind mounted, no open check happens. This is still
  fine, any workload wanting to use the block volume will probably fail if the
  device is not available, and kubelet wouldn't do anything helpful in any
  case.

*  The new "CleanupMountPoint" call is used for the unmount operation


pkg/client/linstor.go
```diff
@@ -1148,55 +1148,53 @@ func (s *Linstor) Mount(vol *volume.Info, source, target, fsType string, readonl
...
 		// This is a block volume so create a file to bindmount to.
 	} else {
-		err := s.mounter.MakeFile(target)
-		if err != nil {
+		f, err := os.OpenFile(target, os.O_CREATE, os.FileMode(0644))
+		if err != nil && !os.IsExist(err) {
 			return fmt.Errorf("could not create bind target for block volume %s, %v", target, err)
 		}
+
+		if f != nil {
```
rck, line 1192: hm, in general I don't like to rely on the "good value" of a function if it returned an error. Usually you can not and should not.

```go
if err != nil {
    if !os.IsExist(err) {...}
} else {
    _ = f.Close()
}
```

rck (changes requested):
besides the 1 comment this LGTM

rck (approved):

PR #126, opened 2021-07-14
Add support for CSI Storage Capacity queries

WanzenBug:
While the `GetCapacity()` call was already implemented, it was not usable in practice:

1. The parameters from the storage class are passed 1:1 to the GetCapacity request. This is an issue when some "external" parameters are present in the storage class. The most obvious example of such a parameter is `csi.storage.k8s.io/fstype`, which we even include in our examples.

   When creating volumes, `csi.storage.k8s.io/fstype` seems to be removed from the request passed to the plugin by the provisioner, but for `GetCapacity`, the parameters seem the remain unchanged. Why that difference exists, I cannot say.

   In any case, to work around this, we ignore parameters with a foreign "namespace" section. For consistency, our own parameters can now also be namespaced using `linstor.csi.linbit.com`. Parameters with no namespace are still treated as "ours" to validate.

2. The internal `CapacityBytes()` call only worked on the parameters, ignoring topology information. The CSI controller will however query the capacity of specific segments (in our case: nodes) in the cluster topology.

   To that end, the call was expanded to include the topology information, which is then used to filter for a specific node and storage pool(s).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   Un-namespaced parameters are still supported, while explicitly namespaced
   parameters that with a foreign namespace are now ignored. These would produce "unknown parameter" errors previously.
 
+### Changed
+
+- Calls to `GetCapacity` now take topology information (if any) into account.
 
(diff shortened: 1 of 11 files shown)


rck (approved):
LGTM

PR #130, opened 2021-08-19
build piraeus-csi for amd64 and arm64

toelke:
See piraeusdatastore/piraeus#74 for the discussion about this feature


toelke:
I have now finally activated `pre-commit`.

toelke:
There needs to be a way to add 🎉 to that "All checks have passed"-message...

I will sign off now for today, you can stop monitoring github :-D

PR #132, opened 2021-09-07
Add new default "AutoPlaceTopology" scheduler

WanzenBug:
This new scheduler is a topology aware version of the old `AutoPlace`
scheduler. Since it is topology aware, it can be used to optimize volume
placement when using `WaitForFirstConsumer` volume binding or restricting
placement via `allowedTopologies`, while still respecting user-defined
placement options such as `replicasOnSame` or `replicasOnDifferent`.

This scheduler was added as all alternatives have some issue:
* `AutoPlace` completely ignored topology information, making it a bad fit for optimized volume placement based on Pod scheduling.
* `FollowTopology` ignored user set constraints  like `replicasOnSame` and `replicasOnDifferent`

### DIFF
--- a/pkg/topology/scheduler/autoplacetopology/autoplacetopology.go
+++ b/pkg/topology/scheduler/autoplacetopology/autoplacetopology.go
@@ -1,0 +1,146 @@
+package autoplacetopology
+
+import (
+	"context"
+	"fmt"
+
+	linstor "github.com/LINBIT/golinstor"
+	lapi "github.com/LINBIT/golinstor/client"
+	"github.com/container-storage-interface/spec/lib/go/csi"
+	"github.com/sirupsen/logrus"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	lc "github.com/piraeusdatastore/linstor-csi/pkg/linstor/highlevelclient"
+	"github.com/piraeusdatastore/linstor-csi/pkg/topology"
+	"github.com/piraeusdatastore/linstor-csi/pkg/topology/scheduler"
+	"github.com/piraeusdatastore/linstor-csi/pkg/volume"
+)
+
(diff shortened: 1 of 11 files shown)


WanzenBug:
Kind of a v2 of #124. While that PR fixed the deviation from the Spec when using `FollowTopology`, it meant that you could either use the Kubernetes Topology XOR Linstor AutoPlace constraints. This new scheduler should address this exact issue.

For example, you might want to have volumes that:
* are scheduled on first use, i.e. with `volumeBindingMode: WaitForFirstConsumer`, which requires topology support
* are distributed over multiple failure domains, requiring LINSTOR autoplace constraints.
Prior to this new default, this wasn't possible.

pkg/topology/scheduler/autoplacetopology/autoplacetopology.go
```diff
@@ -0,0 +1,148 @@
...
+		err := s.Resources.Autoplace(ctx, vol.ID, lapi.AutoPlaceRequest{
+			SelectFilter: lapi.AutoSelectFilter{PlaceCount: 1, NodeNameList: []string{node}},
+		})
+		if err != nil {
+			log.WithError(err).WithField("preferred", node).Trace("failed to autoplace")
+			continue
+		}
+
+		log.WithField("preferred", node).Trace("successfully placed on preferred node")
+
```
rck, line 80: Nitpicking but it puzzled me a bit when seeing that code and finding that it "alwas" `break`s in the happy path. maybe put it in an `else` (to the `err != nil`), or just add a comment like `// successfully autoplaced, we are done`, or maybe just remove the empty line so that it is close to the `log.WithField(... successfully)`. Something like this

rck (approved):
LGTM. There is one minor comment, besides that go for it. :+1:

PR #133, opened 2021-09-27
node: add aux properties as additional topology info

WanzenBug:
In https://github.com/piraeusdatastore/piraeus-operator/commit/530c7aed27ea9640516d40b5095a6eb1248285fe the operator learned to sync node labels to LINSTOR properties. To make CSI (and kubernetes) aware that the linstor plugin is capable of scheduling based on these labels, we need to add them in the node info response.

The change requires some changes in the volume scheduler: now it is possible for a user to specify a general topology, which might not contain the previously assumed single-node identification. Instead of assuming that a segment contains exactly one node, the scheduler now deals with multiple nodes per segment by fetching matching nodes from LINSTOR.

### DIFF
--- a/pkg/client/linstor.go
+++ b/pkg/client/linstor.go
@@ -1459,6 +1459,18 @@ func (s *Linstor) GetNodeTopologies(ctx context.Context, nodename string) (*csi.
 		topo.Segments[label] = "true"
 	}
 
+	node, err := s.client.Nodes.Get(ctx, nodename)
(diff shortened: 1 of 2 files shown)


WanzenBug:
This fixes the last non-passing tests in the current CSI test suite, so hopefully no more reworks for volume scheduling (one can dream, right :innocent: )

rck (approved):
LGTM

PR #136, opened 2021-10-08
get rid of LINSTOR property to pass volume context

WanzenBug:
The CSI Spec includes a way to pass parameters from volume creation down
to publish calls by returning a volume context. Up to now, we achieved this
by adding a magic property, holding a serialized form of the volume
parameters. However this makes it hard to add pre-provisioned volumes
back into kubernetes.

This commit reworks the whole parameter handling, so that:
A) we only use the parameters were strictly required
B) we use the CSI volume context to pass down the parameters.

The only context in which we actually require the parameters is
during node-publish operations (when we format and mount the disk).
This is because the user may have set mkfs and mount options in the storage
class.

In order to support publishing older resources, there is compatibility code
to get that information from the old LINSTOR property.


WanzenBug:
@rck sorry for the big PR, but the `Parameters` struct was used all over, parsed at least thrice on every call and in general a bit of a pain.

Still waiting for some feedback from internal tests, but the code should be good to review.

PR #140, opened 2021-11-04
fix a NPE during snapshot operations

WanzenBug:
golinstor returns nil for the creation timestamp of snapshots in case
linstor does not provide one. This caused an NPE when converting it to
a CSI snapshot object.


rck (approved):
LGTM

PR #142, opened 2021-11-22
support for creating backups to S3 via snapshot mechanism

WanzenBug:
In addition to the normal "in-cluster" snapshots, the user can also select
backups to S3 by providing parameters when creating the snapshot.
    
The parameters can be used to set up a LINSTOR "remote" object, which
represents a S3 endpoint. Any snapshot using those parameters will be
created using LINSTORs backup feature.
    
Restoring from a snapshot that only exists on a remote location does works,
as long as the "remote" is configured beforehand. The CSI spec does not
pass any parameters when registring an existing snapshot.
    
This feature requires LINSTOR 1.18.0 to work. Previous versions did not
allow for customizing the snapshot name. This made it impossible to create
an implementation that adheres to the CSI spec, as there was no way to
map the requested snapshot name to any form of stable snapshot ID.


pkg/client/linstor.go
```diff
@@ -1145,9 +1442,56 @@ func (s *Linstor) ListSnaps(ctx context.Context, start, limit int) ([]*csi.Snaps
...
+		bMap := list.Linstor
+
+		for k := range bMap {
+			if _, ok := foundIds[bMap[k].OriginSnap]; ok {
+				log.WithField("backup", bMap[k].Id).Trace("skipping backup already in cluster")
+				continue
+			}
+
+			if len(bMap[k].Vlms) != 1 {
+				log.WithField("backup", bMap[k].Id).Trace("skipping backup with wrong volume number")
```
JoelColledge, line 1475: As above

pkg/driver/driver.go
```diff
@@ -886,6 +886,15 @@ func (d Driver) CreateSnapshot(ctx context.Context, req *csi.CreateSnapshotReque
 		return nil, missingAttr("CreateSnapshot", req.GetName(), "Name")
 	}
 
+	d.log.WithField("req.parameters", req.GetParameters()).WithField("req.secrets", req.GetSecrets()).Debug("parsing request")
```
JoelColledge, line 889: Are these the secrets themselves? Perhaps best not to log them :smile:

pkg/driver/driver.go
```diff
@@ -886,6 +886,15 @@ func (d Driver) CreateSnapshot(ctx context.Context, req *csi.CreateSnapshotReque
...
+
+	params, err := volume.NewSnapshotParameter(req.GetParameters(), req.GetSecrets())
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "invalid snapshot parameter: %v", err)
```
JoelColledge, line 893:
```suggestion
		return nil, status.Errorf(codes.InvalidArgument, "invalid snapshot parameter: %w", err)
```
Supported here?

pkg/volume/snapshot_params.go
```diff
@@ -0,0 +1,98 @@
...
+
+type SnapshotType int
+
+const (
+	SnapshotTypeInCluster SnapshotType = iota
+	SnapshotTypeS3
+	SnapshotTypeLinstor
+)
+
+type SnapshotParameter struct {
```
JoelColledge, line 25: Singular "Parameter" for the storing the plural "parameters" of the VolumeSnapshotClass?

pkg/volume/snapshot_params.go
```diff
@@ -0,0 +1,98 @@
...
+			}
+
+			p.S3UsePathStyle = b
+		default:
+			log.WithField("key", k).Warn("ignoring unknown snapshot parameter key")
+		}
+	}
+
+	if p.Type != SnapshotTypeInCluster && p.RemoteName == "" {
+		// encode params as json, so we can have a stable hash
```
JoelColledge, line 84: Do we actually want to hash the entire `SnapshotParameter`? Only certain parameters really define the remote. Although, perhaps only `AllowIncremental` is really specific to the snapshot.

JoelColledge (commented):
I don't have much insight into how this fits into the CSI driver as a whole, but the code looks good. Just a bunch of minor points.

pkg/client/linstor.go
```diff
@@ -806,6 +925,130 @@ func (s *Linstor) VolFromSnap(ctx context.Context, snap *csi.Snapshot, vol *volu
...
+		logger.WithError(err).WithField("node", node).Info("failed to restore backup to node")
+	}
+
+	if err != nil {
+		return fmt.Errorf("failed to restore backup to any node (%v), last error: %w", nodes, err)
+	}
+
+	snap, err = s.client.Resources.GetSnapshot(ctx, sourceVolId, snapId)
+	if nil404(err) != nil {
+		return fmt.Errorf("failed to check for presence of local snapshot: %w", err)
```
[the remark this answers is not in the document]
WanzenBug: I like `in-cluster` more. `local` was misleading, but `restored` to me is also misleading, as it might just be a normal snapshot, not a backup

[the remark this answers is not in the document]
JoelColledge: If it were a local snapshot, wouldn't it be found by the first `GetSnapshot` call? Ignoring concurrent snapshot creation, that is. The motivation for my comment was just to differentiate these errors.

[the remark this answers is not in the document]
WanzenBug: Oh, I was looking the the wrong instance of that message :grimacing:

pkg/driver/driver.go
WanzenBug: Seems to use `fmt.Sprintf` internally, so no, i.e. it would just be an `%v`

pkg/driver/driver.go
WanzenBug: Oops. Also logged again a bit below when logging the parsed parameters. Removed the secrets here and added a `Stringer` impl for the parameters that replaces the secret values with `***`.

pkg/client/linstor.go
```diff
@@ -1039,6 +1282,36 @@ func (s *Linstor) reconcileResourcePlacement(ctx context.Context, vol *volume.In
 }
 
 func (s *Linstor) FindSnapByID(ctx context.Context, id string) (*csi.Snapshot, bool, error) {
```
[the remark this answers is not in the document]
WanzenBug: Interface was already documented, copy&pasted it here again

pkg/volume/snapshot_params.go
WanzenBug: Maybe we should just require a name to be specified...

WanzenBug: Replaced by requiring a remote name for S3 snapshots.

JoelColledge (approved):

kvaps:
@WanzenBug am I right that this code uses old "snapshot shipping" mechanism which is considered deprecated?

> The `snapshot ship` command is considered deprecated and any bugs found with it will not be fixed. Instead, use the backup ship command with a remote pointing to your local controller. For more details, see [the previous section](https://linbit.com/drbd-user-guide/linstor-guide-1_0-en/#s-shipping_snapshots-l2l).

kvaps:
It seems it uses new backups API

https://github.com/WanzenBug/linstor-csi/blob/85a424eba39108d575414a40014d87fdb2ef6e32/pkg/client/linstor.go#L771

WanzenBug:
Yeah, it only uses the new backups API.

PR #143, opened 2021-11-23
Advanced volume access rules

WanzenBug:
Up to now, using diskless volumes was an all-or-nothing affair: either allow
diskless resource from any node, or don't allow them at all. This was quite
limiting for more advanced cluster architectures.

For example, clusters spanning multiple zones could not restrict diskless
access to only nodes in the same zone. Similarly, if you had dedicated storage
hosts, you could not restrict diskless volumes to a specific subset.

Another issues was in the calculation of storage capacity. In case of dedicated
storage hosts, the storage capacity was effectively 0 on almost all nodes,
meaning pods were never scheduled on those nodes, even tough diskless access
from the dedicated storage was possible.

As a solution, introduce new values for the `allowRemoteVolumeAccess`
parameter. It now takes a list of matching rules, that are compared to the
topology segments provided by the CO.

For example, to allow access from nodes in the same zone, you could now use:

```yaml
allowRemoteVolumeAccess: |
  - fromSame:
    - topology.kubernetes.io/zone
```

For access from anywhere, you can still use the old "true", which desugars to

```yaml
allowRemoteVolumeAccess: |
  - fromSame: []
```

The old strict access "false" desugars to:

```yaml
allowRemoteVolumeAccess: |
  - fromSame:
    - linbit.com/hostname
```

### DIFF
--- a/pkg/volume/parameter.go
+++ b/pkg/volume/parameter.go
@@ -76,7 +76,7 @@ type Parameters struct {
 	// Encrypt volumes if true.
 	Encryption bool
 	// AllowRemoteVolumeAccess if true, volumes may be accessed over the network.
-	AllowRemoteVolumeAccess bool
+	AllowRemoteVolumeAccess RemoteAccessPolicy
 	// LayerList is a list that corresponds to the `linstor resource create`
 	// option of the same name.
 	LayerList []devicelayerkind.DeviceLayerKind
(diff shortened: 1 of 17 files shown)


pkg/volume/parameter.go
```diff
@@ -76,7 +76,7 @@ type Parameters struct {
 	// Encrypt volumes if true.
 	Encryption bool
 	// AllowRemoteVolumeAccess if true, volumes may be accessed over the network.
-	AllowRemoteVolumeAccess bool
```
rck, line 79: so that is a breaking change or did I overlook something?

WanzenBug: Not really. `true` and `false` are now syntax sugar (see commit message). Observable behaviour should remain the same

rck: right, that was me overlooking the `MarshalText/UnmarshalText`

rck (approved):
fingers crossed, but LGTM

PR #144, opened 2021-12-02
attach: ignore replicasOn... parameters if required

WanzenBug:
replicasOn... parameters are intended for spreading the intial set of replicas.
For volume use, it might be required to attach the volume to a node that would
violate the requirements. As a high level argument, you could say that these
temporary attached volumes are not really replicas in the sense implied by
replicasOn... parameters.

The fix is to fall back to manual resource creation if required. Note that we
do not always use this manual creation, as in some cases it is advantageous
to try make-available first. For example, in a shared storage pool scenario
make-available will pick a new shared resource over diskless attachment.

### DIFF
--- a/pkg/client/linstor.go
+++ b/pkg/client/linstor.go
@@ -424,58 +424,14 @@ func (s *Linstor) Attach(ctx context.Context, volId, node string, readOnly bool)
 		return err
 	}
 
-	// In certain circumstances it is necessary to create a diskfull resource to make it usable.
-	// The bug report that introduced this variable is a good example:
-	// * Our cluster has 4 identical nodes called A, B, C, D.
-	// * A resource was placed in a typical 2 + 1 Tiebreaker configuration, let's say the diskfull resources are
-	//   on A and B, the Tiebreaker on C.
-	// * A Pod attaches on node A.
-	// * Node A goes down/becomes unreachable/etc. That leaves us with 1 diskfull resource on B + 1 diskless on C.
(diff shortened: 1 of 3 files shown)


pkg/client/linstor.go
```diff
@@ -486,18 +442,69 @@ func (s *Linstor) Attach(ctx context.Context, volId, node string, readOnly bool)
...
+		// The bug report that introduced this variable is a good example:
+		// * Our cluster has 4 identical nodes called A, B, C, D.
+		// * A resource was placed in a typical 2 + 1 Tiebreaker configuration, let's say the diskfull resources are
+		//   on A and B, the Tiebreaker on C.
+		// * A Pod attaches on node A.
+		// * Node A goes down/becomes unreachable/etc. That leaves us with 1 diskfull resource on B + 1 diskless on C.
+		// * The Pod is deleted and a replacement scheduled on node D (i.e. after deletion by the HA Controller).
+		// * Now, this method will be called, since we need to Attach on node D, and there is currently no resource on that
+		//   node.
+		// * Using a diskless resource will not work, as the newly creates resource would not get quorum.
```
rck, line 455: `s/creates/created/`

rck (approved):
hm, can't say it makes me particularly happy to see this "we do it like this! but when the moon is in the first quarter we do it differently" kind of code, that makes it hard to reason about paths taken and simply introduces multiple paths/combinations of paths that need to be tested and thought of. For now as a quick solution I don't see anything else, so "ack" as in "I don't have to maintain it, if I would I probably would not take it".

WanzenBug:
Yeah, sadly that whole attach logic is a mess.

PR #145, opened 2021-12-16
Update CSI Spec + dependencies

WanzenBug:
* CSI Spec 1.5.0
* Replace `go:generate` commands with go1.17 `go run <module>@<version>` syntax.
* Update dependencies (requires go 1.17)

See the commit messages for more information


pkg/client/mocks/BackupProvider.go
```diff
@@ -30,17 +30,24 @@ func (_m *BackupProvider) Abort(ctx context.Context, remoteName string, request
 }
 
 // Create provides a mock function with given fields: ctx, remoteName, request
-func (_m *BackupProvider) Create(ctx context.Context, remoteName string, request client.BackupCreate) error {
+func (_m *BackupProvider) Create(ctx context.Context, remoteName string, request client.BackupCreate) (string, error) {
```
rck, line 33: why did this happen in this commit?

rck (commented):
looks good overall, I just wonder about what triggered that one change in the middle commit

pkg/client/mocks/BackupProvider.go
WanzenBug: I updated golinstor, which changed the signature for that function to return the snapshot name. So I had to rerun the mock generation

rck: ah, that was it. thanks.

PR #146, opened 2022-01-10
New "usePvcName" parameter for user-controlled names

WanzenBug:
A new parameter `usePvcName` is added. When "true", the driver will (try to)
use the PVC name + namespace instead of the generated PV name as ID. Because
this might generate name conflicts in certain scenarios, it remains disabled
by default.

### DIFF
--- a/pkg/client/linstor.go
+++ b/pkg/client/linstor.go
@@ -250,7 +250,17 @@ func (s *Linstor) FindByID(ctx context.Context, id string) (*volume.Info, error)
 	}), nil
 }
 
-func (s *Linstor) CompatibleVolumeId(name string) string {
+func (s *Linstor) CompatibleVolumeId(name, pvcNamespace, pvcName string) string {
+	if pvcNamespace != "" && pvcName != "" {
+		s.log.Debug("try creating valid volume name based on PVC")
+
+		generatedName := fmt.Sprintf("%s-%s", pvcNamespace, pvcName)
+
+		if validResourceName(generatedName) == nil {
+			return generatedName
+		}
+	}
+
 	invalid := validResourceName(name)
 	if invalid == nil {
 		return name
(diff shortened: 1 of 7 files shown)


pkg/driver/driver.go
```diff
@@ -487,7 +487,18 @@ func (d Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
...
 
-	volId := d.Storage.CompatibleVolumeId(req.GetName())
+	params, err := volume.NewParameters(req.GetParameters())
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "failed to parse parameters: %v", err)
+	}
+
+	var pvcNamespace, pvcName string
+	if params.UsePvcName {
+		pvcName = req.GetParameters()[ParameterCsiPvcName]
```
rck, line 497: the values for the keys for sure exist? (and are "" or some defined good value)?

pkg/client/linstor.go
```diff
@@ -250,7 +250,18 @@ func (s *Linstor) FindByID(ctx context.Context, id string) (*volume.Info, error)
...
 }
 
-func (s *Linstor) CompatibleVolumeId(name string) string {
+func (s *Linstor) CompatibleVolumeId(name, pvcNamespace, pvcName string) string {
+	if pvcNamespace != "" && pvcName != "" {
+		s.log.Debug("try creating valid volume name based on PVC")
+
+		generatedName := fmt.Sprintf("%s-%s", pvcNamespace, pvcName)
+
+		err := validResourceName(generatedName)
```
rck, line 259: could be a `if err := ....`

rck (commented):
LG in general, 2 minor comments. Also the commit msg looks weird? "use the PVC name + namespace instead +  of the generated PV"

pkg/driver/driver.go
WanzenBug: They don't have to exist. But that should then just default to empty strings, in which case they get ignored again :shrug:

rck: that would be fine, I was just worried that it might access keys that don't exist at all

rck (approved):
LGTM now, feel free to merge after the checks succeeded

PR #147, opened 2022-01-13
unpublish: ignore errors generated by deleted nodes

WanzenBug:
If a node is deleted from k8s and LINSTOR, the volume attachments might
still hang around, triggering ControllerUnpublishVolume calls on nodes
that no long exist.

This commit ensures we handle this case gracefully, detaching from the "lost"
node. Since the node was already deleted from LINSTOR, there isn't actually
anything for us to-do on detach.


WanzenBug:
Note: draft until e2e tests have passed

rck (approved):
LGTM, feel free to merge when the e2e tests pass

PR #148, opened 2022-01-17
respect parameters when restoring from snapshots

WanzenBug:
It's possible to restore from a snapshot while changing the parameters
of the restored volume. For example, you can use different placement count,
placement policy, etc... Up to now this information was completely ignored.
A snapshot was restored on the same nodes as the original resource, without
regard to placement count or policy.

This commit adds logic to restore from a snapshot while changing placement
and placement count.

For a start, only one node is selected to restore from. The remaining nodes
to fulfil the placement count are selected based on the normal scheduling
priority, as would be the case with new volumes.

Also change the scheduling logic slightly to deal with partially deployed
resources when selecting a preferred node. When restoring, it is possible
that one node has a the restored resource, but it is not one of the preferred
nodes. Now the scheduler tries to first place on the preferred nodes,
regardless of the currently deployed resources.


rck (approved):
It does not get easier :). but LGTM

PR #157, opened 2022-04-01
delegate mkfs to LINSTOR

WanzenBug:
Replaces custom mkfs formatting code with LINSTOR built-ins. This replaces
the formatAndMount check at node attach time with simply setting properties
on the ResourceDefinition.

Note that we do not set it on the ResourceGroup (~ StorageClass) level, as
filesystem is determined by the PersistentVolumeClaim (~ ResourceDefinition)
being created.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,7 +10,7 @@ repos:
         args:
           - --multi
       - id: check-added-large-files
-  - repo: git://github.com/dnephin/pre-commit-golang
+  - repo: https://github.com/dnephin/pre-commit-golang
     rev: v0.4.0
     hooks:
       - id: go-fmt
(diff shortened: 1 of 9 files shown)


rck (commented):
LGTM and feel free to merge as is. The only thing I wonder about is if it slightly changes the semantic (even if in a better way). So far what you wrote is what you got, right. And now LINSTOR might add some options that are in general a good idea on DRBD, AFAIK. What if one disables these options here because (bad idea, but you get the point)? All of that comes with using "LINSTOR flavored mkfs" which is good, but should we somehow document it more prominently? Just a sentence in the `CANGELOG.md`

WanzenBug:
Thanks, I honestly didn't know that LINSTOR adds some option on their own (it's basically the `nodiscard` option for thin volumes currently). And I also noticed that upgrading CSI might lead to unusable volumes. These volumes by definition can't hold any relevant data, so I think that is fine to ignore.

PR #158, opened 2022-04-06
Fix mtab warning for mkfs commands

kvaps:
Docker image `debian:bullseye-slim` does not contain `/etc/mtab` symlink. While docker is setting it [automatically](https://github.com/moby/moby/blob/master/daemon/initlayer/setup_unix.go#L25-L36) for every container, containerd is not doing this.

This affects Kubernetes deployments with containerd, which are report the following warning every time when attaching a new volume to pod:
```
mke2fs 1.44.5 (15-Dec-2018)
ext2fs_check_if_mount: Can't check if filesystem is mounted due to missing mtab file while determining whether /dev/drbd1000 is mounted.
```

This confuses the end users of linstor-csi plugin.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,7 +28,8 @@ ARG LINSTOR_WAIT_UNTIL
 
 RUN apt-get update && apt-get install -y --no-install-recommends \
       xfsprogs e2fsprogs \
-      && apt-get clean && rm -rf /var/lib/apt/lists/*
+      && apt-get clean && rm -rf /var/lib/apt/lists/* \
+      && ln -sf /proc/mounts /etc/mtab
 
 COPY --from=builder /linstor-csi /
 COPY --from=downloader /linstor-wait-until /linstor-wait-until


kvaps:
Should I create a similar PR for [piraeus-server](https://github.com/piraeusdatastore/piraeus/blob/master/dockerfiles/piraeus-server/Dockerfile) image?

WanzenBug (approved):
LGTM. I guess the same is needed for piraeus-server, although I don't think the users get's to see the log message there, unless trace logging is enabled?

kvaps:
**linstor-csi**: this error is shown every time when any other error occurs with linstor-csi.
**piraeus-server**: this error is shown only when user manually exec into linstor-node pod and run `mkfs.ext4 /dev/drbd1000` there. If I remember correctly. linstor-satellites are never run mkfs commands by themself.

WanzenBug:
> linstor-satellites are never run mkfs commands by themself.

That's changed with https://github.com/piraeusdatastore/linstor-csi/commit/1ffbefb8f4ae74012f3b04855649fb1c4e3041d2

PR #161, opened 2022-05-04
order snapshot restore based on preferred nodes

WanzenBug:
When restoring a snapshot from S3 backup, LINSTOR CSI will have to select
a node on which to download the backup. This should have always been the
preferred node (which was likely chosen by the workload scheduler). However,
we only considered the requisite topology, which is not always good enough.

With this commit we sort the requisite nodes by their preference, so in the
ideal case, the snapshot is downloaded directly on the node that will consume
the new volume.


rck (approved):
LGTM

Issue #155, opened 2022-03-16
Volume attachments fail when Quorum is disabled past version 0.13.1


tw185112:
Expected/Previous behaviour: 
- Our setup:
	- Linstor Cluster with Quorum feature disabled.
	- Cluster is running on the edge and is limited to 2 diskful nodes. This cannot be changed as it is a limitation in our product.
	- Redundant pair of servers.
- We are using Piraeus to provide storage to Kubernetes cluster where nodes are running diskless
- Version 0.11.0 was working fine:
	- When one of the two diskful nodes was offline, we create a pod which mounts a Persistent Volume stored in Linstor,
		the volume is attached successfully and storage can be accessed in the volume.
		When the failed diskful node is restored, DRBD restores the storage. We have accepted the risk of split brain and have
		mitigations for this.
- Resource group (Kubernetes) in Linstor define place count as 2.

Current/Observed behaviour:
- Volume attachment no longer succeeds when updated to 0.13.1 version.
- The following 404 error is produced from the csi-controller - `msg="method failed" func="github.com/sirupsen/logrus.(*Entry).Error" file="/go/pkg/mod/github.com/sirupsen/logrus@v1.4.2/entry.go:297" error="rpc error: code = Internal desc = ControllerPublishVolume failed for pvc-fa6e974e-1eee-4750-af4e-20c06a01e5e5: 404 Not Found"`
- The Linstor-plugin  produces the error (quorum error) - `msg="1 replicas of 2 are apparently not reachable, create a new diskfull resource for quorum"`
- This is unexpected because we are running with quorum disabled.
- Upon further investigation, we believe this error is caused by the [this line ](https://github.com/piraeusdatastore/linstor-csi/blob/v0.13.1/pkg/client/linstor.go#L471) in the Linstor-CSI code.

We would be happy to collab to find a solution for this issue.
Potential options we have identified:
- Return to original behaviour (regression possible for other situations).
- Insert a way to turn off that check, if the controller is configured to run with quorum disabled i.e. to skip these checks
- Alternatively, use a config map to configure to run in a quorum free mode, so skip quorum checks. (Configuration of the pod is changed).

We have rolled back to 0.11.0 for now but would like to fix to move forward with updates in the future.

WanzenBug:
Thanks for the detailed explanation and investigation. This is definitly unexpected behaviour, mostly because we _strongly_ encourage everyone to use quorum whenever possible.

[...]

PR #163, opened 2022-05-12
ISSUE #155: useQuorum parameters added to check if quorum is enabled/…

tw185112:
…disabled to stop volume attachments from failing

Signed-off-by: Tom Wilson <thomas.wilson@ncr.com>

### DIFF
--- a/pkg/client/linstor.go
+++ b/pkg/client/linstor.go
@@ -227,6 +227,7 @@ func (s *Linstor) resourceDefinitionToVolume(resDef lapi.ResourceDefinitionWithV
 		ResourceGroup: resDef.ResourceGroupName,
 		FsType:        fsType,
 		Properties:    props,
+		UseQuorum:     resDef.Props["DrbdOptions/Resource/quorum"] != "off",
 	}
 }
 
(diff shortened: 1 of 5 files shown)


pkg/client/linstor.go
```diff
@@ -488,10 +495,24 @@ func (s *Linstor) Attach(ctx context.Context, volId, node string, readOnly bool)
...
+			return nil
+		} 
+		for k, v := range controllerProps {
+			if k == "DrbdOptions/auto-quorum" {
+				if v != "disabled" {
+					controllerQuorum = true
+				}
+			}
+		}
+		useQuorum := volQuorum || controllerQuorum
```
WanzenBug, line 511: We don't need to check the controller properties here. 

First of all we would also need to check the resource group property.

Secondly, this setting does not influence _existing_ resources, so only whatever is set directly on the resource definition is relevant.

So: just remove this and use `volQuorum` directly.

pkg/client/linstor.go
```diff
@@ -227,6 +233,7 @@ func (s *Linstor) resourceDefinitionToVolume(resDef lapi.ResourceDefinitionWithV
 		ResourceGroup: resDef.ResourceGroupName,
 		FsType:        fsType,
 		Properties:    props,
+		UseQuorum:     quorum,
```
WanzenBug, line 236:
```suggestion
		UseQuorum:     resDef.Props["DrbdOptions/Resource/quorum"] != "off",
```

WanzenBug (changes requested):
Please see that https://github.com/piraeusdatastore/linstor-csi/pull/163/checks pass.

Also some cleanup/simplification is needed. The logic change itself seems sound.

WanzenBug (approved):
LGTM. I'll run some e2e tests on this, but I don't expect any issues.

rck:
@WanzenBug you squash it before merging, right?

WanzenBug:
@tw185112 thank you for your contribution!

PR #167, opened 2022-06-03
snapshots: only snapshots if nodes are online

### LINKED ISSUE #166: snapshots: check that nodes are online before taking snapshots
If a node is offline during snapshot creation, LINSTOR creates an error report. Since there is no good back-off strategy implemented for the snapshot controller, that means we create a lot of error reports. We should probably check that involved nodes are online before performing the snapshot creation.

WanzenBug:
snapshots with offline nodes create errors in LINSTOR. Since there is no
good back-off strategy employed by who-ever creates the snapshot requests,
we create linstor errors in a tight loop. To mitigate this issue, we check
the resource state before snapshotting. If LINSTOR does not report any
state, we can assume the node to be offline.

Also requires an upgrade of golinstor, which means regenerating the mocks.

Closes #166


CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+- A new checks ensures snapshots are only attempted when the resource is completely in sync. This prevents an issue
```
rck, line 11: singular/plural

rck (approved):
besides the minor spelling mistake the code LGTM

PR #169, opened 2022-06-10
CSI 1.6.0

WanzenBug:
* Update CSI to 1.6.0
* Implement volume status reporting for controller
  * Report "published" nodes, i.e. where a resource is available
  * Report DRBD status if possible


rck (approved):
LGTM. Wanted to complain what a horrible name "abnormal" is, then saw the struct field in `csi.VolumeCondition`...

The only thing that came to my mind was if status querying could be too expensive? How often does that happen, can it be a problem for the controller? But that is something you know better, so if this isn't a problem go ahead and merge.

WanzenBug:
> The only thing that came to my mind was if status querying could be too expensive?

Definitly something to keep an eye on. I'll do some tests internally, but I think right now those methods are basically never called.

WanzenBug:
Actually, it is never called, unless we deploy https://github.com/kubernetes-csi/external-health-monitor/ Monitor interval seems to be 5 minutes, which is relatively tame, I think

PR #176, opened 2022-10-05
remove quorum workaround

WanzenBug:
The CSI driver had a special case for attaching volumes to a degraded cluster. This is no longer necessary, as DRBD now (9.1.12/9.2.0) calculates quorum differently, meaning adding diskless to a degraded cluster still works.

### DIFF
--- a/pkg/client/linstor.go
+++ b/pkg/client/linstor.go
@@ -447,7 +447,7 @@ func (s *Linstor) GetLegacyVolumeParameters(ctx context.Context, volId string) (
 }
 
 // Attach idempotently creates a resource on the given node.
-func (s *Linstor) Attach(ctx context.Context, volId, node string, readOnly, rwxBlock, volQuorum bool) error {
+func (s *Linstor) Attach(ctx context.Context, volId, node string, readOnly, rwxBlock bool) error {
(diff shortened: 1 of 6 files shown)


pkg/client/linstor.go
```diff
@@ -611,28 +570,17 @@ func (s *Linstor) Attach(ctx context.Context, volId, node string, readOnly, rwxB
...
-func inspectExistingResource(resource *lapi.Resource) (*bool, string) {
+// getDisklessFlag inspects a resource to determine the right diskless flag to use.
+func getDisklessFlag(resource *lapi.Resource) string {
 	layer := &resource.LayerObject
 
-	isAvailableDiskful := true
-	isUnavailableDiskful := false
-
 	for layer != nil {
 		if layer.Type == devicelayerkind.Drbd {
```
rck, line 578: as this got simpler, make the `layer.Type` a switch? As is these are just 2 weird independent `if`s. At least an `else if` I'd say.

rck (commented):
all in all LGTM, just 1 comment about a bit of code cleanup

pkg/client/linstor.go
WanzenBug: Thanks, changed to a switch.

Issue #159, opened 2022-04-06
kubevirt: allow-two-primaries must be set to perform live-migration


kvaps:
To reproduce:

Install KubeVirt and enable featureGates: `LiveMigration` and `HotplugVolumes`

```yaml
apiVersion: kubevirt.io/v1
kind: KubeVirt
metadata:
  annotations:
    kubevirt.io/latest-observed-api-version: v1
    kubevirt.io/storage-observed-api-version: v1alpha3
  name: kubevirt
  namespace: kubevirt
spec:
  configuration:
    developerConfiguration:
      featureGates:
      - LiveMigration
      - HotplugVolumes
  imagePullPolicy: IfNotPresent
```

Create VM:
```yaml
apiVersion: kubevirt.io/v1
kind: VirtualMachine
metadata:
  annotations:
    kubevirt.io/latest-observed-api-version: v1
    kubevirt.io/storage-observed-api-version: v1alpha3
  name: testvm1
  namespace: default
spec:
  running: true
  template:
    metadata:
      creationTimestamp: null
    spec:
      domain:
        devices:
          disks:
          - disk:
              bus: virtio
            name: containerdisk
          - disk:
              bus: virtio
            name: cloudinitdisk
          interfaces:
          - masquerade: {}
            name: default
        machine:
          type: q35
        resources:
          requests:
            memory: 1024M
      networks:
      - name: default
        pod: {}
      terminationGracePeriodSeconds: 30
      volumes:
      - containerDisk:
          image: kubevirt/fedora-cloud-container-disk-demo:latest
        name: containerdisk
      - cloudInitNoCloud:
          userData: |-
            #cloud-config
            password: fedora
            chpasswd: { expire: False }
        name: cloudinitdisk
```

Create volume:
```yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: my-pvc
spec:
  accessModes:
    - ReadWriteMany
  volumeMode: Block
  storageClassName: linstor-data-r2
  resources:
    requests:
      storage: 100Gi
```

attach volume to the vm:
```
virtctl addvolume testvm --volume-name my-pvc
```

run live-migration:
```
virtctl migrate testvm1
```

You'll see that hotplug-volume pod will be failed to run on second node:

```
hp-volume-29d84               1/1     Running             0          2m4s
hp-volume-6lvvm               0/1     ContainerCreating   0          77s
```

Due to:
```
Events:
  Type     Reason                  Age               From                     Message
[...]

PR #177, opened 2022-11-02
Support live-migration with allow-two-primaries

kvaps:
This PR allows to live-migrate KubeVirt virtual machines by setting `allow-two-primaries` flag dynamically.

fixes https://github.com/piraeusdatastore/linstor-csi/issues/159

## Changes in this PR:

* **controllerPublishVolume:**
    * if volume has `type: Block` and `accessMode: ReadWriteMany`;
    * if it is already `InUse` at one other node;
    
   enables `allow-two-primaries` flag

* **controllerUnpublishVolume:**
  * if volume `InUse` at one one less amount of nodes

  removes `allow-two-primaries` flag

### DIFF
--- a/pkg/client/linstor.go
+++ b/pkg/client/linstor.go
@@ -447,7 +447,7 @@ func (s *Linstor) GetLegacyVolumeParameters(ctx context.Context, volId string) (
 }
 
 // Attach idempotently creates a resource on the given node.
-func (s *Linstor) Attach(ctx context.Context, volId, node string, readOnly, volQuorum bool) error {
+func (s *Linstor) Attach(ctx context.Context, volId, node string, readOnly, rwxBlock, volQuorum bool) error {
 	s.log.WithFields(logrus.Fields{
 		"volume":     volId,
 		"targetNode": node,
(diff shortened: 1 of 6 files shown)


WanzenBug (changes requested):
Change looks good, just need to fix `pkg/client/linstor_test.go` to use the extra argument on `Attach()`

WanzenBug:
@kvaps If you want, I can take care of the tests + lints

kvaps:
Sure, tests are fixed

WanzenBug (approved):
I'll just quickly start the e2e test suite, then we can merge this

pkg/client/linstor.go
```diff
@@ -636,6 +653,30 @@ func (s *Linstor) Detach(ctx context.Context, volId, node string) error {
 		"targetNode": node,
 	})
 
+	ress, err := s.client.Resources.GetAll(ctx, volId)
+	if err != nil {
```
WanzenBug, line 657:
```suggestion
	if nil404(err) != nil {
```

Otherwise we fail the CSI tests because a "double" delete will result in an error

WanzenBug: Or actually just `return nil404(err)` in the branch

WanzenBug:
Thanks!

PR #181, opened 2022-12-21
update golinstor

WanzenBug:
Use golinstor based initialization of LINSTOR client. This enables checking the server TLS without requiring a mutual TLS config, as well as using the new built-in User-Agent and Bearer Token configs.


rck (approved):
LGTM

PR #186, opened 2023-02-24
Add _netdev option to all mounts

WanzenBug:
This ensures unmounting happens before the network stack is torn down during node shutdown, preventing situations where DRBD would lose quorum before systemd had the chance to unmount all CSI controlled mounts.


WanzenBug:
Note to self: requires mounting `/run/mount` from the host.

rck (approved):
LGTM

PR #188, opened 2023-03-23
ignore unschedulable nodes in capacity requests

WanzenBug:
When scheduling Pods, the default kubernetes scheduler takes into account the reported storage capacity of nodes. Up to now, we also reported the capacity of evicted/evacuated or "AutoplaceTarget=false" nodes. This meant that Kubernetes tried to schedule Pods on nodes where storage could never be attached.

The (partial) fix is to ignore nodes during storage capacity accounting. This means users with stricter topology requirements get a reported capacity of 0 for those nodes.

It is not a complete fix, as from the CSI side we don't have a way mark the node as completely unscheduable.

### DIFF
--- a/pkg/client/linstor.go
+++ b/pkg/client/linstor.go
@@ -730,6 +730,33 @@ func (s *Linstor) CapacityBytes(ctx context.Context, storagePool string, segment
 		return 0, fmt.Errorf("unable to get capacity: %w", err)
 	}
 
+	allNodes, err := s.client.Nodes.GetAll(ctx, &lapi.ListOpts{Cached: &cached})
+	if err != nil {
+		return 0, fmt.Errorf("unable to get nodes: %w", err)
+	}
+
+	var filteredNodes []string
+
+	for i := range allNodes {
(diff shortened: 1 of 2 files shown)


pkg/client/linstor.go
```diff
@@ -730,6 +730,33 @@ func (s *Linstor) CapacityBytes(ctx context.Context, storagePool string, segment
...
+
+		if !slice.ContainsString(requestedNodes, node.Name) {
+			continue
+		}
+
+		if slice.ContainsString(node.Flags, lapiconsts.FlagEvicted) || slice.ContainsString(node.Flags, lapiconsts.FlagEvacuate) {
+			continue
+		}
+
+		if node.Props[lapiconsts.KeyAutoplaceAllowTarget] == "false" {
```
rck, line 751: not in this commit, but if we have that more often some helper that gets the slice and a key and checks for "does it exist" and if yes, is it false (for whatever string version of false) might make sense.

rck (approved):
LGTM

PR #196, opened 2023-04-27
Fix multiple requisites

kvaps:
fixes https://github.com/piraeusdatastore/linstor-csi/issues/195


kvaps:
It seems driver works fine without it

PR #198, opened 2023-05-04
Use property namespace to prefix replica-on-* conf

WanzenBug:
Since 0.21.0, users where able to change the default "namespace" where the CSI driver would expect topology information to be available on the node.

Until now, we did not change the prefix functionallity for settings replicas-on-same/different in the parameters: they were always prefixed with Aux/.

This became a problem with Operator v2, which configures Aux/topology as the default namespace for these kinds of labels. To fix, we change the prefix function to include the support the property namespace.

All "absolute" property names (i.e. those already prefixed with Aux/) are left untouched, so people can still refer to properties outside the namespace if they wish to do so.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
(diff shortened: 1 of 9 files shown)


rck (approved):
guess you want to squash that, besides that I think this is fine.

PR #199, opened 2023-06-01
remove PUBLISH_READONLY capability

WanzenBug:
The capability is intended for systems that can ensure read-only mode when publishing on the controller, i.e. when we would create a diskless resource if necessary. We do not support a "global" RO flag in LINSTOR: it is only the local decision of the mounting node.

Additionally, the readOnly flag on the ControllerPublishVolume request remained unused by Kubernetes, another reason to remove it.

This makes it safe to remove it, cleaning up some code along they.

As an additional improvement, automatically detect the readOnly flag for mounts involving the MULTI_NODE_READER_ONLY mode.


rck (approved):
code LGTM. I just can not reason if we are taking away a real feature or if that is "dead code" anyways. If you feel we don't take away something actually useful to anyone, go ahead.

PR #201, opened 2023-06-12
implement deletion for backups

WanzenBug:
Implement deletion for S3 snapshots/backups by calling the appropriate API in the DeleteSnapshot() call.

In order to speed up this process, we also enhance the existing Find/ListSnapshot functions to return the originating LINSTOR remote name. This both identifies the item as a backup, and helps with the delete calls.


WanzenBug:
@rck ping :)

rck (approved):
LGTM

PR #202, opened 2023-06-15
implement deletion of local snapshots for S3 backups

WanzenBug:
When creating a remote backup in LINSTOR, we can remove the local snapshot after shipping completes.

### DIFF
--- a/pkg/driver/driver.go
+++ b/pkg/driver/driver.go
@@ -956,8 +956,14 @@ func (d Driver) CreateSnapshot(ctx context.Context, req *csi.CreateSnapshotReque
 		}).Debug("found existing snapshot")
 
 		if existingSnap.GetSourceVolumeId() == req.GetSourceVolumeId() {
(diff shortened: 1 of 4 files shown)


pkg/driver/driver.go
```diff
@@ -967,6 +976,14 @@ func (d Driver) CreateSnapshot(ctx context.Context, req *csi.CreateSnapshotReque
 		return nil, status.Errorf(codes.Internal, "failed to create snapshot: %v", err)
 	}
 
+	if snap.ReadyToUse && params.Type != volume.SnapshotTypeInCluster && params.DeleteLocal {
```
rck, line 979: this whole block is basically a copy from above. please make that a function; at least the check.

rck (changes requested):

pkg/driver/driver.go
WanzenBug: Good point, moved to a function and cleaned it up a bit.

rck (approved):
LGTM

PR #208, opened 2023-07-12
Prevent shrinking of volumes

WanzenBug:
This is normally enforced by the provisioner, but when restoring from a backup without reported snapshot size, the provisioner cannot enforce this limit.

Instead we simply update the check to ensure the volume definition is never shrunk.


rck (approved):
LGTM

PR #212, opened 2023-10-06
support deletion of snapshots after restore from backup

WanzenBug:
When a user requests a restore from an off-site backup, they might want the local snapshot to be deleted after the restore.

We already have this option when creating the snapshot by using a special parameter on the snapshot class. However, we do not have that information available directly.

To support this, we need to lookup the snapshot content and snapshot class ourselfes, then parse the parameters. This is all optional, as LINSTOR CSI still needs to work outside kubernetes, or inside Kubernetes with incomplete information.

### DIFF
--- a/pkg/driver/driver.go
+++ b/pkg/driver/driver.go
@@ -39,6 +39,11 @@ import (
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/rest"
 
 	"github.com/piraeusdatastore/linstor-csi/pkg/client"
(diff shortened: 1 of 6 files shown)


pkg/driver/driver.go
```diff
@@ -1259,7 +1279,12 @@ func (d Driver) createNewVolume(ctx context.Context, info *volume.Info, params *
 					"CreateVolume failed for %s: snapshot not found in storage backend", req.GetName())
 			}
 
-			if err := d.Snapshots.VolFromSnap(ctx, snap, info, params, req.GetAccessibilityRequirements()); err != nil {
+			snapParams, err := d.maybeGetSnapshotParameters(ctx, snap)
+			if err != nil {
+				logger.WithError(err).Warn("failed to fetch snapshot parameters, continuing without it")
+			}
+
+			if err := d.Snapshots.VolFromSnap(ctx, snap, info, params, snapParams, req.GetAccessibilityRequirements()); err != nil {
```
rck, line 1287: this feels a bit weird. you might get an `err`, that in this case is not fatal, so you only log it. usually in go if you have `x, err := f()` and you get an `err`, you can not assume anything about `x`. But you then feed `snapParams` into `VolFromSnap`. That just feels wrong, maybe set `snampParams` to `nil`, or to some "defined value" in case of an `err`?

rck (commented):
LGTM, feel free to merge

PR #229, opened 2023-11-09
Add over provision factor

WanzenBug:
Adds a new overprovision parameter that calculates the capacity of a given
    storage pool based on the existing reserved space.
    
While we would like to use LINSTOR directly for that, it does not seem to
    be implemented: the "query-size-info" call seems to return inconsistent
    results.
    
The advantage of the new parameter is that people can opt-in to this new
    calculation by setting the parameter.
    
This also adds a cache for the call to `Resources.GetResourceView()`, as the CapacityBytes() is one of the "hottest" code paths, and we don't actually need to monitor the usage that closely.

### DIFF
--- a/pkg/linstor/highlevelclient/high_level_client.go
+++ b/pkg/linstor/highlevelclient/high_level_client.go
@@ -25,6 +25,7 @@ import (
 	"sync"
 	"time"
 
+	lc "github.com/LINBIT/golinstor"
 	lapi "github.com/LINBIT/golinstor/client"
 	"github.com/container-storage-interface/spec/lib/go/csi"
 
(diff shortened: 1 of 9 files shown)


pkg/linstor/highlevelclient/high_level_client.go
```diff
@@ -251,30 +260,80 @@ func (n *NodeCacheProvider) GetStoragePoolView(ctx context.Context, opts ...*lap
...
 	for _, item := range items {
-		if _, ok := filterNames[getNodeName(item)]; ok {
-			result = append(result, item)
+		if len(filterPools) > 0 {
+			if _, ok := filterNames[getNodeName(item)]; !ok {
+				continue
+			}
+		}
+
+		if len(filterPools) > 0 {
```
rck, line 330: both ifs check the `filterPools`?

rck (commented):
1 comment and in general: hm, and that can not be changed or fixed in LINSTOR?

WanzenBug:
> that can not be changed or fixed in LINSTOR?

Yes and no. There are multiple issue along the way:
* The `query-size-info` call returns unexpected results, not taking into account the overprovision factor
* When `CapacityBytes` is called, there is not necessarily a RG available
* The way the results are generated by `query-size-info` (i.e. available space on each node) is not compatible with the way the information is queried by LINSTOR CSI: limiting the result to just one node is not possible AFAIK

In general, I would like to see this moved to LINSTOR, but I fear it will take a LINSTOR v2 to get around to it. Perhaps then we also have an AutoPlacer that supports all the requirements one can already specify in LINSTOR CSI :-)

PR #230, opened 2023-11-14
backups: select random node to download backup

WanzenBug:
When chosing a node to download our backup, we should prevent the same node from downloading the backups all the time.

In the case we get a "preferred" topology, this is left to the kubernetes scheduler/CSI controllers: they should select different preferred nodes.

In the case we have no preferred topology given to use, we should do our own random selection. Otherwise we end up with the same node being chosen to download backups all the time, while the others sit idle.


phoenix-bjoern:
LGTM. Thanks @WanzenBug !

rck (approved):
LGTM

PR #243, opened 2024-01-24
Fixes around attaching volumes with unusual resources

WanzenBug:
Fixes some issues in the AttachVolume call and beyond:
* Do not try to attach a resources that does not have a replica
* Do not try to create a diskless resource if the resource does not support it
* Set default access policy to "local only" for resources without replication layer


rck (approved):
LGTM

PR #252, opened 2024-03-13
params: allow setting multiple storage pools

WanzenBug:
LINSTOR Resource Groups can have multiple storage pools set in a single resource group. LINSTOR then selects one matching pool for every resource placement.

Since storage pool names cannot contain any spaces, there should be no existing storage class with spaces in the storage pool parameter. This means we can simply reuse the existing parameter.


rck (approved):
LGTM, currently a bit busy with other things. Approved, feel free to merge if you are confident enough :)

PR #256, opened 2024-03-29
delay deletion of volume definitions

WanzenBug:
We delete the volume definition in order to have resource definitions without resources. This is useful when we want to keep snapshots without having the associated resources around.

During the volume deletion, we first delete the volume definition. This may be causing a bug in DRBD 9.2, so we delay the deletion until after all resources are gone, at which point DRBD has fully shut down the resource.


rck (approved):
LGTM

PR #261, opened 2024-04-23
use golinstor caching instead of custom implementation

WanzenBug:
Reuse the caching implementation from golinstor, it is probably more complete and less error prone.


rck (approved):
LGTM

PR #283, opened 2024-08-07
examples: update k8s deployment

WanzenBug:
* Reuse the resource we already configure for the Operator
* Update instructions on how to use these deployments
* Small fixes to the readme


rck (approved):
LGTM

PR #294, opened 2024-10-04
do not try to resize block volumes

WanzenBug:
Block volumes cannot be resized, so there is no point in even attempting it. An attempt would fail because it could not determine the FS one the volume, as there likely isn't any. A retry would then succeed as there is an earlier check if we need to mount at all.

The fix is to rearrange the logic to only attempt to resize filesystem volumes.

While at it, also remove the deprecated "IsNotMountPoint". The double negative was getting confusing in any case.


rck (approved):
LGTM

PR #297, opened 2024-10-25
snapshot: do not restore on offline nodes

WanzenBug:
A node may be offline. Since we select only one node, we should ensure it is not offline.

### DIFF
--- a/pkg/client/linstor.go
+++ b/pkg/client/linstor.go
@@ -1317,13 +1317,25 @@ func (s *Linstor) reconcileSnapshotResources(ctx context.Context, snapshot *volu
 
 	// Optimize the node we use to restore. It should be one of the preferred nodes, or just the first with a snapshot
 	// if no preferred nodes match.
-	selectedNode := snap.Nodes[0]
-
+	var selectedNode string
 	for _, snapNode := range snap.Nodes {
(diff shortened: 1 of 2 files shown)


pkg/client/linstor.go
```diff
@@ -1317,15 +1317,27 @@ func (s *Linstor) reconcileSnapshotResources(ctx context.Context, snapshot *volu
...
+		}
+
+		if selectedNode == "" {
+			selectedNode = snapNode
+		}
+
 		if slice.ContainsString(preferredNodes, snapNode) {
 			selectedNode = snapNode
 			break
 		}
```
rck, line 1334: could you make the first `if` part of an `else` in the second `if`, I think this makes it easier to read. as in

```go
if slice.ContainsString(...) {
    // we found a perfect candidate
    ...
    break
} else if selectedNode == "" {
   // better than nothing
   ....
}
```

rck (changes requested):
LGTM in general, I guess restructuring the code makes it a bit easier to read

pkg/client/linstor.go
WanzenBug: I agree. Updated based on your suggestion

rck (approved):
LGTM

PR #309, opened 2024-12-23
disable fallback to RO mounts for RW volumes

WanzenBug:
If a user requests a RW volume, we disable the default behaviour of "mount" to fallback to RO mounts should there be an issue with the initial RW attempt.

This may happen if a DRBD device is in a weird state, where the default auto-promote does not succeed, but the resource is otherwise fine. A RO mount does not require an explicit state change, so it may succeed in such a situation.

By requesting "rw" in the mount options, we prevent this behaviour.


rck (approved):
LGTM

PR #312, opened 2024-12-31
really disable fallback mounting to RO

WanzenBug:
Passing the "rw" option is not enough. Careful reading of the manpage, verified by tests and reading of the source code of util-linux reveals that disabling the RO fallback only happens when you pass the "-w" flag.

Then, also fix "ro" mounts in case of XFS, as XFS attempts to write even if passed the "ro" flag without the "norecovery" option.

Fixes: 6deb565 ("disable fallback to RO mounts for RW volumes")

### DIFF
--- a/pkg/client/linstor.go
+++ b/pkg/client/linstor.go
@@ -27,6 +27,7 @@ import (
 	"math"
 	"math/rand"
 	"os"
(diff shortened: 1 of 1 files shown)


rck (approved):
LGTM

PR #320, opened 2025-02-18
Set FS type from LINSTOR properties during mounts

WanzenBug:
If a user restores a volume from a snapshot, the file system type of the restored volume may not match the requested FS type of the storage class. This may happen if for example the snapshot was taken from an xfs volume, but the new StorageClass is configured to provision ext4 volumes. We then would try to call "mount -t ext4" on a xfs-formatted block device, leading to a failure mode where the volume is created fine, but you cannot make use of it.

To fix this, instead of taking the FS type directly from Kubernetes, we use the existing LINSTOR FsType property to determine what fs type is actually on the volume. Only if this information is not provided (possible with very old snapshots or some other funky setups), we fall back to the information provided by Kubernetes.

One alternative would be to disallow such mixing of filesystem types completely, ensuring when a volume is restored from a snapshot that old and new volume have matching FS type in Kubernetes. This would permanently lock in some users to their initial choice of FS, so we instead chose the option that allows for gradual migration to a new StorageClass/FS.


rck (approved):
looks like a sound decision, code also LGTM

PR #333, opened 2025-04-30
better volume validation

WanzenBug:
Up to now, we did not properly validate volume capabilities during volume creation, allowing users to create FileSystem volumes with RWX capabilities.

Careful reading of the CSI standard shows that these capabilities need to be checked during CreateVolume and ValidateVolumeCapabilities calls.


rck (approved):
LGTM

PR #340, opened 2025-06-03
Add automatic remote configuration for snapshot classes

WanzenBug:
When the CSI driver is asked to create a new volume from a snapshot, the snapshot must be "known" already. However, in the case that a new cluster was provisioned, and some tool populated the snapshot objects from scratch, the CSI driver is never given the information in the VolumeSnapshotClass.

This means that in a new cluster, where the user configured the SnapshotClass for some LINSTOR S3/L2L Remote, the CSI driver will fail because it has no access to the remotes.

To fix this, we need to manually watch for VolumeSnapshotClasses and try to keep them in sync with the remotes configured in LINSTOR.


rck (approved):
this looks equally plausible as horrible :)

PR #343, opened 2025-06-24
fix issues with left-over resource definition after volumes have been deleted

WanzenBug:
* golinstor used to invalidate the cache before applying modifications. This might cause stale cache entries if the cache gets re-populated before the modification is made. The end result was that sometimes the CSI driver thought that certain resource should not be deleted. An update to golinstor fixes this issue.
* LINSTOR periodically runs a job that recreates replicas based on the RG settings. This sometimes interfered with our deletion logic, causing resource to not be properly cleaned up, leaving resources with no volumes and the resource definition behind. To fix this, disable the task for the resources we intent to delete.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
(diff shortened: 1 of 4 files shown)


rck (approved):
LGTM

PR #345, opened 2025-07-01
enforce larger minimum volume size for FS volumes

WanzenBug:
Some automations might cause the creation of requests for 10MiB FS volumes. For example, this may happen in virtualization settings with cloud-init volumes and similiar scenarios.

This caused issues, as ext4 and xfs have their own minimum volume size. For XFS, it is 300MiB, for ext4 it is possible to create tiny volumes, but below 2MiB journal creation is disabled, which for "HA" is a bad idea.


rck (approved):
LGTM

PR #349, opened 2025-07-10
supported LINSTOR volume clones

WanzenBug:
We already supported cloning volumes via snapshots. Since LINSTOR now can also clone volumes, and be smart about mixing and matching different storage pools, we can switch to using the LINSTOR clone path, enabling clones on more types of storage pools.

### DIFF
--- a/pkg/client/linstor.go
+++ b/pkg/client/linstor.go
@@ -35,6 +35,7 @@ import (
 
 	lapiconsts "github.com/LINBIT/golinstor"
 	lapi "github.com/LINBIT/golinstor/client"
+	"github.com/LINBIT/golinstor/clonestatus"
 	"github.com/LINBIT/golinstor/devicelayerkind"
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	"github.com/haySwim/data"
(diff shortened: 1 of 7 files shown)


pkg/client/linstor.go
```diff
@@ -372,6 +378,101 @@ func (s *Linstor) Create(ctx context.Context, vol *volume.Info, params *volume.P
...
+
+	logger.Debugf("check for clone status")
+
+	status, err := s.client.ResourceDefinitions.CloneStatus(ctx, src.ID, vol.ID)
+	if err != nil && !errors.Is(err, lapi.NotFoundError) {
+		logger.Debugf("clone status failed: %v", err)
+		return err
+	}
+
+	if errors.Is(err, lapi.NotFoundError) {
```
rck, line 403: this is a bit weird and "jumpy". maybe do it like

```go
if err != nil {
  if errors.Is(err, lapi.NotFoundError) {
    ...
  } else {
    logger...
    return err
  }
}
```

or the other way around to have the 2 liner in the true part. something like that

pkg/client/linstor.go
```diff
@@ -372,6 +378,101 @@ func (s *Linstor) Create(ctx context.Context, vol *volume.Info, params *volume.P
...
+	_, err = s.reconcileVolumeDefinition(ctx, vol)
+	if err != nil {
+		logger.Debugf("reconcile volume definition failed: %v", err)
+		return err
+	}
+
+	logger.Debug("reconcile volume placement")
+
+	err = s.reconcileResourcePlacement(ctx, vol, params, topologies)
+	if err != nil {
```
rck, line 460: personally for these cases I prefer `if err := call(); err != nil {` but that is probably just a matter of taste, so feel free to ignore

rck (changes requested):
one comment that is probably worth simplifying the code a bit, and one to ignore

PR #364, opened 2025-09-18
Support for RWX volumes via NFS

WanzenBug:
This adds the necessary infrastructure to dynamically provision and use RWX filesystem volumes with LINSTOR CSI. There is a new image `nfs-server` that exports volumes as NFS, which then get mounted by LINSTOR CSI.

The details are laid out in docs.

For review I recommend working through the commits one by one.

### DIFF
--- a/cmd/linstor-csi/linstor-csi.go
+++ b/cmd/linstor-csi/linstor-csi.go
@@ -24,6 +24,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"strings"
 	"time"
 
 	linstor "github.com/LINBIT/golinstor"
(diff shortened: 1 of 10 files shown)


nfs/Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+ARG TARGETOS
+ARG VERSION=unknown
+RUN --mount=type=cache,target=/root/.cache/go-build GOOS=$TARGETOS GOARCH=$TARGETARCH CGO_ENABLED=0 \
+    go build \
+    -a \
+    -ldflags "-X github.com/piraeusdatastore/linstor-csi/pkg/driver.Version=$VERSION -extldflags -static"  \
+    -o /nfs-helper  \
+    ./cmd/nfs-helper
+
+FROM debian:bookworm
```
rck, line 19: trixie?

nfs/Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+
+RUN { echo 'APT::Install-Recommends "false";' ; echo 'APT::Install-Suggests "false";' ; } > /etc/apt/apt.conf.d/99_piraeus
+
+RUN --mount=type=cache,target=/var/cache,sharing=private \
+    --mount=type=cache,target=/var/lib/apt/lists,sharing=private \
+    --mount=type=tmpfs,target=/var/log \
+    apt-get update && \
+    apt-get install -y wget ca-certificates && \
+    wget https://packages.linbit.com/public/linbit-keyring.deb -O /var/cache/linbit-keyring.deb && \
```
rck, line 28: does not matter too much, especially not for debian, but whenever possible this should use a distro specific keyring from https://packages.linbit.com/public/linbit-keyring/

cmd/nfs-helper/start-stop-reactor.go
```diff
@@ -0,0 +1,144 @@
...
+			return
+		}
+
+		if err := os.Rename(temp, dest); err != nil {
+			log.Printf("failed to rename '%s' to '%s': %s\n", temp, dest, err)
+			return
+		}
+	}
+
+	for _, f := range previousFiles {
```
rck, line 133: guess you would not need that. you could read all files here and if it it not in `cm.Data` it is a previous file. makes the code a bit simpler, but also potentially reads more files. does not really matter in the end I guess

nfs/Dockerfile
WanzenBug: Thanks! Did not know about https://packages.linbit.com/public/linbit-keyring/ :sweat_smile:

PR #381, opened 2025-11-04
run fsck before mounting

WanzenBug:
We had repeated issues with ext4 in particular, as that expects to run e2fsck from time to time. Why a journaling filesystem is so paranoid remains unknown, but we can try to be nice and run fsck.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Run "fsck" before mounting a filesystem
+
 ## [1.10.0] - 2025-10-14
 
 ### Added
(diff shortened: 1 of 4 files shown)


BoleynSu:
This breaks RWO when multiple pods mount the same pvc on the same node.
```
  Warning  FailedMount             11s (x7 over 43s)  kubelet                  MountVolume.SetUp failed for volume "pvc-d9539098-6f5e-4aec-88ce-8f8b4553afd4" : rpc error: code = Internal desc = NodePublishVolume failed for pvc-d9539098-6f5e-4aec-88ce-8f8b4553afd4: failed to run fsck on device '/dev/drbd1000': failed to run fsck: output: "fsck from util-linux 2.41
/dev/drbd1000 is mounted.
e2fsck: Cannot continue, aborting.
```

@rck @WanzenBug

rck:
wasn't that fixed in https://github.com/piraeusdatastore/linstor-csi/pull/386 ?

PR #395, opened 2025-12-07
Fix: include node name in topology segments for RemoteAccessPolicyLoc…

chrstnwhlrt:
GenericAccessibleTopologies was not including the node hostname in the segments map passed to AccessibleSegments(). This caused RemoteAccessPolicyLocalOnly (allowRemoteVolumeAccess: false) to return an empty topology, effectively disabling nodeAffinity on PVs.

The fix adds topology.LinstorNodeKey to the segments map, allowing RemoteAccessPolicyLocalOnly to properly restrict volume access to only the nodes where diskful replicas exist.

### DIFF
--- a/pkg/linstor/highlevelclient/high_level_client.go
+++ b/pkg/linstor/highlevelclient/high_level_client.go
@@ -76,6 +76,11 @@ func (c *HighLevelClient) GenericAccessibleTopologies(ctx context.Context, volId
 			}
 		}
 
+		// Include the node name in segments so that RemoteAccessPolicyLocalOnly
+		// (which uses topology.LinstorNodeKey) can properly restrict access to
+		// only the nodes where the volume has diskful replicas.
+		segs[topology.LinstorNodeKey] = nodes[i].Name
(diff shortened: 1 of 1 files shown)


chrstnwhlrt:
You're right – this is specific to external LINSTOR clusters not managed by Piraeus Operator.

In our setup, nodes only have Aux/dc set for zone-awareness, but no Aux/linbit.com/hostname property:

$ linstor node list-properties node-f-0
Aux/dc = falkenstein
NodeUname = node-f-0

Without Aux/linbit.com/hostname, AccessibleSegments() returns an empty map, causing GenericAccessibleTopologies() to return nil – so PVs get no nodeAffinity.

The fix ensures RemoteAccessPolicyLocalOnly works regardless of whether the hostname is set via Aux/ properties (Operator-managed) or derived from the node name (external clusters).

Alternatively, we could document that external clusters must set Aux/linbit.com/hostname on all nodes – but that seems error-prone and easy to forget when adding new nodes.

PR #397, opened 2025-12-12
fix compatibility with LINSTOR 1.33 backup API

WanzenBug:
LINSTOR 1.33.0 changed the backup API, so that a successful up- or download
    is no longer indicated by a flag on the snapshot resource. Instead some
    properties are used in order to support shipping to multiple targets.
    
    This meant our existing implementation that checks if a LINSTOR snapshot is
    actually a "remote" snapshot broke. To fix this, we first check for the new
    properties on the snapshot and use those as basis for the snapshot remote
    names and readiness.
    
    LINSTOR now allows shipping to more than one target, but we only expect one
    "remote" location to exist, we need to select the correct one. In our case,
    this should always be the earliest remote that started shipping, so if there
    are multiple possible destination, we report the one that first started
    uploading.

In a previous commit, I replaced the custom `slice` package with the one from go stdlib.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Compatibility with LINSTOR 1.33+ when sending snapshots to S3.
+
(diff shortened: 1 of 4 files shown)


rck (approved):
LGTM

PR #429, opened 2026-04-03
fix: prevent resource definition deletion when PVC still exists

kvaps:
## Summary

Before deleting a resource definition, check if the corresponding Kubernetes PV
is still Bound to a PVC that is not being deleted. This prevents accidental RD
deletion when resources are stuck in DELETING state.

## Problem

When DRBD resources get stuck in DELETING state (e.g. due to bitmap mismatch
after toggle-disk), `deleteResourceDefinitionAndGroupIfUnused()` checks only
whether all resources have the `FlagDelete` flag. If they do, it proceeds to
delete the entire resource definition -- even if the PVC is still active and
Bound.

This can lead to data loss: the PVC remains Bound, the VM continues running
with cached data, but the LINSTOR resource definition is deleted.

## Fix

Add a safety check at the beginning of `deleteResourceDefinitionAndGroupIfUnused()`:
query the Kubernetes API for the PV (by name, which matches the LINSTOR resource
name). If the PV exists, is not being deleted, and is Bound to a PVC that is also
not being deleted -- abort the RD deletion.

Uses the existing `dynamic.Interface` client pattern already used elsewhere in the
codebase.

### DIFF
--- a/cmd/linstor-csi/linstor-csi.go
+++ b/cmd/linstor-csi/linstor-csi.go
@@ -36,6 +36,7 @@ import (
 	"github.com/piraeusdatastore/linstor-csi/pkg/client"
 	"github.com/piraeusdatastore/linstor-csi/pkg/driver"
 	lc "github.com/piraeusdatastore/linstor-csi/pkg/linstor/highlevelclient"
+	"github.com/piraeusdatastore/linstor-csi/pkg/utils"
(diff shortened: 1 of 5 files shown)


JoelColledge:
Thanks for looking into this. There are commits in this PR that are unrelated to the description. Please keep the PR focused on one topic.

WanzenBug:
This will be replaced by the new "resource truncate" API in LINSTOR 1.35.

PR #435, opened 2026-04-27
feat(attach): force Protocol C on dual-attach for Protocol-A/B resources

kvaps:
## Problem

DRBD rejects `allow-two-primaries=yes` on resources configured with Protocol A or B. `drbdadm adjust` fails with `Protocol C required` (errno 139), and any live migration of a KubeVirt VM whose volume sits on a Protocol-A/B resource ends up in a permanent evacuation loop.

This is a real-world setup: operators commonly opt into Protocol A on a per-resource-group basis (e.g. a "replicated-async" StorageClass) for async / WAN replication. That choice silently breaks every subsequent live migration of consumers of those volumes.

## Change

When `allow-two-primaries=yes` is installed on the resource-definition during a second Attach, also install `DrbdOptions/Net/protocol=C` as an override on the resource-definition itself — but only when the resource currently effectively uses Protocol A or B.

The override is set at the resource-definition level (rather than per resource-connection) so it applies to every connection — including connections to diskless peers like a TieBreaker, where a per-pair override would still leave the satellite's adjust failing on the unaltered connection.

A small `Aux/csi-protocol-override` marker is set alongside the protocol override so Detach can safely revert exactly the override we installed, without disturbing a Protocol property an operator may have set on the resource-definition themselves. Once `resInUse == 1` again, the override is dropped and the resource falls back to its resource-group default.

## Compatibility

- No behaviour change for resources that already use Protocol C (the common case).
- No behaviour change for resources that are never attached with allow-two-primaries.
- Idempotent: re-running Attach is a no-op once override is installed.
- Operator-set Protocol overrides on the resource-definition are preserved (gated by the Aux marker).

## Tests

- `go build ./...` clean
- `go vet ./pkg/...` clean
- `go test ./pkg/client/...` passes
- Verified end-to-end on a 3-node cluster (KubeVirt v1.6.3): VM with a PVC backed by a Protocol-A resource-group now live-migrates successfully in a single Attach, with the override installed during dual-attach and removed after Detach. Without the patch the same Attach fails on the satellite with `Protocol C required` and KubeVirt loops on evacuation.

### DIFF
--- a/pkg/client/linstor.go
+++ b/pkg/client/linstor.go
@@ -653,11 +653,35 @@ func (s *Linstor) Attach(ctx context.Context, volId, node string, rwxBlock bool)
 	}
 
 	if otherResInUse > 0 && rwxBlock {
-		rdPropsModify := lapi.GenericPropsModify{OverrideProps: map[string]string{
+		rdProps := map[string]string{
 			linstor.PropertyAllowTwoPrimaries: "yes",
-		}}
+		}
 
-		err = s.client.ResourceDefinitions.Modify(ctx, volId, rdPropsModify)
+		// DRBD requires Protocol C whenever allow-two-primaries is enabled.
+		// If the resource is configured with Protocol A or B (typically
+		// through its resource-group), drbdadm adjust on the satellites
+		// fails with "Protocol C required" (errno 139) and live migration
+		// cannot proceed. Override Protocol to C on the resource-definition
+		// itself so it applies to every connection (including diskless
+		// peers, e.g. a TieBreaker). The marker lets Detach revert exactly
+		// the override we installed without disturbing operator-set props.
+		proto, protoErr := s.getEffectiveDrbdProtocol(ctx, volId)
+		if protoErr != nil {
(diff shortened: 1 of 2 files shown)


WanzenBug (commented):
That is indeed a tricky corner case. The idea seems sound, but I am wondering:

1. I believe it would be simpler to unconditionally apply the override. I don't think we need to expect it to be set another way. LINSTOR CSI would use the RG properties from the storage class, and the Operator does not mess with RD properties at all. If a user manually overwrote it: tough look, we don't really support that anywhere. Then we can simply skip the check for what the current protocol is and apply this RD property always.
2. I'm not sure how these properties interact with `linstor node-connections`. Those would be the way a user should configure per-node settings for replication (for example, using protocol A between nodes in different regions). Maybe setting this on the RD already overwrites this, but you would have to try. If not, this might still fail because the applied protocol is still A or B.

WanzenBug:
This is superseeded by LINSTOR 1.35, which, when using `make-available --auto-manage-dual-primary` will set protocol C for the "right" connections.

PR #437, opened 2026-05-04
remove published nodes capability

WanzenBug:
The CSI spec tells us:

> A list of all `node_id` of nodes that the volume in this entry
> is controller published on.
> This field is OPTIONAL. If it is not specified and the SP has
> the LIST_VOLUMES_PUBLISHED_NODES controller capability, the CO
> MAY assume the volume is not controller published to any nodes.
> If the field is not specified and the SP does not have the
> LIST_VOLUMES_PUBLISHED_NODES controller capability, the CO MUST
> not interpret this field.
> published_node_ids MAY include nodes not published to or
> reported by the SP. The CO MUST be resilient to that.

We recently changed the logic from reporting all nodes that have potential access, listing all nodes with the resource on it, to just those with explicit diskless attachment.

This breaks in hyperconverged setup, because now the csi-attacher believes the node was externally unpublished when it is attached on an existing diskful node.

Since there is no easy way for us to track the "publishing" state within LINSTOR, we just remove the capability all together: it does not have any actual use for us.

Fixes: 9f0c867540cc ("fix(driver): use CSI property to identify published nodes in ListVolumes")
Closes #410


rck (approved):
quite frankly I can't tell if this makes a difference, but the code LGTM

PR #439, opened 2026-05-05
disable provisioning of RWX volumes without DRBD layer

WanzenBug:
RWX requires NFS controlled by DRBD Reactor, so we always need a DRBD layer for RWX volumes. For block volumes, we do not use NFS, but still without any replication, there is no use for RWX.

### DIFF
--- a/pkg/driver/driver.go
+++ b/pkg/driver/driver.go
@@ -36,6 +36,7 @@ import (
 	"time"
 
 	lc "github.com/LINBIT/golinstor"
+	"github.com/LINBIT/golinstor/devicelayerkind"
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	"github.com/sirupsen/logrus"
 	"google.golang.org/grpc"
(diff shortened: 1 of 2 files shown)


pkg/driver/driver.go
```diff
@@ -599,6 +600,14 @@ func (d Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 		volumeBytes[1] = requiredBytes
 	}
 
+	if len(params.LayerList) > 0 && params.LayerList[0] != devicelayerkind.Drbd {
```
rck, line 603: checking for `len()` and possibly skipping the rest of the check is fine here? can the len be 0? if so, do we then get the DRBD layer from "somewhere else" so that we don't need to check here? or is len of 0 an error in itself?

WanzenBug: `len() == 0` would either cause an error from LINSTOR, or LINSTOR to apply the default "drbd storage" layer. I guess either case is fine for us here :sweat_smile:

PR #440, opened 2026-05-06
Monitor NFS export startup

WanzenBug:
There have been reports of exports failing because ganesha somehow starts but does not create the right socket. So add a basic check that runs after the NFS server starts and verifies that it properly came up.

To be honest: I do not like this. This seems like a horrible hack to work around an issue that should not exist. But apparently, NFS servers are fickle beasts :/


rck (approved):
LGTM

PR #445, opened 2026-05-31
fix(utils): match RWX PV lookup by CSI volumeHandle

immanuwell:
follow-up to #439

Bug
RWX block validation looks up the PV by `volumeID` as if that was the PV object name.
If `metadata.name` and `spec.csi.volumeHandle` differ, the check just misses the PV and bails. Kinda sneaky.

Fix
Fall back to listing PVs and match by `spec.csi.volumeHandle`.
Add a regression test for that case too.

Repro
1. Create a CSI PV where `metadata.name` is not equal to `spec.csi.volumeHandle`.
2. Bind it to a PVC used by 2 pods from different KubeVirt VMs.
3. On `master`, `ControllerPublishVolume` skips the RWX check, so the bad attach can slip through.
4. With this patch, the driver finds the PV by `volumeHandle` and returns `FailedPrecondition`.

Test
`go test ./...`

### DIFF
--- a/pkg/utils/rwx_validation.go
+++ b/pkg/utils/rwx_validation.go
@@ -227,11 +227,9 @@ func findPersistentVolumeByVolumeID(ctx context.Context, kubeClient dynamic.Inte
 
 	for i := range pvList.Items {
 		volumeHandle, found, nestedErr := unstructured.NestedString(pvList.Items[i].Object, "spec", "csi", "volumeHandle")
-		if nestedErr != nil || !found || volumeHandle != volumeID {
-			continue
-		}
-
-		return &pvList.Items[i], nil
+	if volumeHandle == volumeID {
+	     return &pvList.Items[i], nil
+	}
 	}
 
 	return nil, fmt.Errorf("persistentvolume with volumeHandle %q not found", volumeID)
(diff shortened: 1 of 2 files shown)


pkg/utils/rwx_validation.go
```diff
@@ -211,6 +212,31 @@ func ValidateRWXBlockAttachment(ctx context.Context, kubeClient dynamic.Interfac
...
+		return nil, err
+	}
+
+	for i := range pvList.Items {
+		volumeHandle, found, nestedErr := unstructured.NestedString(pvList.Items[i].Object, "spec", "csi", "volumeHandle")
+		if nestedErr != nil || !found || volumeHandle != volumeID {
+			continue
+		}
+
+		return &pvList.Items[i], nil
```
WanzenBug:
```suggestion
	if volumeHandle == volumeID {
	     return &pvList.Items[i], nil
	}
```

I believe this is nicer to read

WanzenBug (changes requested):

WanzenBug (changes requested):
Please ensure the commit is signed off and fix the errors

WanzenBug:
Made obsolete by #474

PR #448, opened 2026-06-03
fix: reclaim incremental S3 backups instead of failing to delete them

### LINKED ISSUE #324: Snapshot Deletion Errors with Velero and Linstor’s Snapshot Shipping
Issue moved over from https://github.com/LINBIT/linstor-server/issues/441

---

I’ve been experimenting with Velero and Linstor’s snapshot shipping feature. Initially, I was confused about the configuration. It turns out that Velero’s data mover must be disabled, or else both Linstor and Velero’s Kopia start shipping snapshot data to S3. I believe this may cause a race condition, potentially leading to [this issue](https://github.com/LINBIT/linstor-server/issues/440).

This requirement doesn’t seem to be clearly documented, so it might be worth adding a note about it somewhere. Perhaps in the article "[Backing up a Kubernetes Deployment Using Velero & LINBIT SDS](https://linbit.com/blog/using-velero-linbit-sds-to-back-up-restore-a-kubernetes-deployment/)"

## Problem: Snapshot Deletion Errors
Once I disabled snapshotMoveData, I wanted to test how incremental snapshots are handled. I created a backup schedule to take a snapshot every 2 minutes, with a retention of 10 minutes:

**Backup schedule**
```yaml
apiVersion: velero.io/v1
kind: Schedule
metadata:
  name: sample-schedule
  namespace: velero
spec:
  schedule: "*/2 * * * *"
  template:
    includedNamespaces:
      - default
    includedResources:
      - '*'
    storageLocation: default
    ttl: "0h10m0s"
    # snapshotMoveData: true
```

Initially, everything worked fine, but once the first snapshots started expiring, the CSI snapshot container began throwing errors.

**Example error log:**
`
E0303 22:00:41.509762       1 snapshot_controller_base.go:361] could not sync content "snapcontent-4e7c6cdd-a156-41e0-a1b9-03516af60b1e": failed to delete snapshot "snapcontent-4e7c6cdd-a156-41e0-a1b9-03516af60b1e", err: failed to delete snapshot content snapcontent-4e7c6cdd-a156-41e0-a1b9-03516af60b1e: "rpc error: code = Internal desc = unable to delete snapshot snapshot-4e7c6cdd-a156-41e0-a1b9-03516af60b1e: failed to delete backup pvc-b5ca1c80-f58b-4aa6-bd9f-279ee235ebc2_back_20250303_215030^snapshot-4e7c6cdd-a156-41e0-a1b9-03516af60b1e: Message: 'pvc-b5ca1c80-f58b-4aa6-bd9f-279ee235ebc2_back_20250303_215030^snapshot-4e7c6cdd-a156-41e0-a1b9-03516af60b1e.meta should be deleted, but at least pvc-b5ca1c80-f58b-4aa6-bd9f-279ee235ebc2_back_20250303_215231^snapshot-00255afe-6945-4035-a04a-28b4b55cd21c.meta is referencing it. Use --cascading to delete recursively'"
`

For full details, see the [csi-snapshotter.log](https://github.com/user-attachments/files/19163370/csi-snapshotter.log).

**Snapshot Class Configuration**
```yaml
apiVersion: snapshot.storage.k8s.io/v1
kind: VolumeSnapshotClass
metadata:
  name: piraeus-snapshots
  labels:
    velero.io/csi-volumesnapshot-class: "true"
driver: linstor.csi.linbit.com
deletionPolicy: Retain
parameters:
  snap.linstor.csi.linbit.com/type: S3
  snap.linstor.csi.linbit.com/remote-name: velero-backup
  snap.linstor.csi.linbit.com/allow-incremental: "true"
  snap.linstor.csi.linbit.com/delete-local: "false"
  snap.linstor.csi.linbit.com/s3-bucket: **redacted**
  snap.linstor.csi.linbit.com/s3-endpoint: **redacted**
  snap.linstor.csi.linbit.com/s3-signing-region: **redacted**
  snap.linstor.csi.linbit.com/s3-use-path-style: "true"
  csi.storage.k8s.io/snapshotter-secret-name: linstor-csi-s3-access
  csi.storage.k8s.io/snapshotter-secret-namespace: velero
```

**Question**
- Are these errors expected when using incremental snapshots in combination with Velero, or could this be a bug?
- Could it be that Velero continuously attempts to delete snapshots, leading to these errors?

**Bonus Question**
With Velero, it’s common to have multiple backup schedules with varying backup intervals and TTLs. For example:

```text
Interval: Hourly    ttl: 72 hours
Interval: Daily     ttl: 7 days
Interval: Weekly    ttl: 4 weeks
Interval: Monthly   ttl: 4 months
```

How does Linstor’s CSI handle such cases?
   - When does it decide to create a full backup versus an incremental backup?
   - If multiple schedules target the same volume, how does it determine which snapshots to retain or delete?

boedy:
## Problem

Closes #324.

With `allow-incremental: "true"` on an S3 `VolumeSnapshotClass`, LINSTOR ships a full backup followed by a chain of incrementals. When an old backup expires the driver is asked to delete it, but LINSTOR refuses because it is the *base* of a later incremental:

```
... should be deleted, but at least ... is referencing it. Use --cascading to delete recursively
```

The delete never succeeds, so nothing in the chain is ever reclaimed — reported downstream as unbounded snapshot growth.

## Changes

**1. Defer deletion of a still-referenced base.** `SnapDelete` now lists all of the resource's backups and, if the one being deleted is still the base of another increment (`based_on_id`), returns a retryable error instead of failing. The
external-snapshotter retries, so the chain drains newest-first with no orphaned data and no cascading delete of increments we want to keep. (`based_on_id` carries a `.meta` suffix the backup id does not, which is stripped before comparing.)

**2. Bound the chain (opt-in).** Two new parameters force a fresh full backup so old chains become reclaimable under rolling retention:

| Parameter | Meaning |
|---|---|
| `snap.linstor.csi.linbit.com/max-increments` | start a new full once the chain has this many increments (unset/`0` = no limit) |
| `snap.linstor.csi.linbit.com/full-snapshot-after` | force a full once the chain's full backup is older than this duration |

Without this, a retained newer increment pins the chain back to the original full and nothing is reclaimed.

## Testing

- Unit tests for leaf detection (incl. the `.meta` suffix), chain-length calculation (incl. in-progress/failed backups), and parameter parsing.
- Manually verified against LINSTOR + MinIO: reproduced the issue, confirmed the chain drains to zero with the fix, and exercised both bounding parameters and incremental restore.

## Open questions

- Default for `max-increments`: opt-in (as here) vs. a sensible default-on value so a stock incremental schedule self-reclaims — §1 alone does not bound steady-state growth.
- Deferral signal: retryable error (as here) vs. another mechanism so the `VolumeSnapshotContent` is not held open during the overhang.

## Compatibility

Requires a LINSTOR backup API that populates `based_on_id`. No new RBAC, CRDs, or dependencies; behaviour is unchanged for full-only backups and any backup with no dependents.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- `snap.linstor.csi.linbit.com/max-increments` and
+  `snap.linstor.csi.linbit.com/full-snapshot-after` snapshot class parameters to
+  bound the length / age of incremental S3 backup chains, forcing a fresh full
+  backup so that older backups become reclaimable.
+
+### Fixed
+
+- Deletion of incremental S3 backups no longer fails permanently when a backup
+  is still the base of a newer increment. The delete is deferred and retried so
+  the chain drains newest-first instead of leaking backups forever.
+
 ## [1.11.2] - 2026-05-12
 
 ### Changed
--- a/examples/k8s/volume-snapshot-class.yaml
+++ b/examples/k8s/volume-snapshot-class.yaml
@@ -18,6 +18,12 @@ parameters:
   # Delete local copy of the snapshot after uploading completes
   snap.linstor.csi.linbit.com/delete-local: "true"
   snap.linstor.csi.linbit.com/allow-incremental: "false"
+  # When incremental backups are enabled, bound the chain so that older backups
+  # eventually become reclaimable. Start a fresh full backup once the chain has
+  # this many increments (omit or 0 = no limit)...
+  # snap.linstor.csi.linbit.com/max-increments: "10"
+  # ...and/or once the chain's full backup is older than this duration.
+  # snap.linstor.csi.linbit.com/full-snapshot-after: "168h"
   snap.linstor.csi.linbit.com/s3-bucket: snapshot-bucket
   snap.linstor.csi.linbit.com/s3-endpoint: s3.us-west-1.amazonaws.com
   snap.linstor.csi.linbit.com/s3-signing-region: us-west-1
--- a/pkg/client/linstor.go
+++ b/pkg/client/linstor.go
@@ -1033,10 +1033,27 @@ func (s *Linstor) reconcileBackup(ctx context.Context, id string, params *volume
 		}
 
 		if info == nil {
+			incremental := params.AllowIncremental
+
+			// Bound the chain length/age when requested: forcing a fresh full
+			// starts a new chain so older backups eventually become reclaimable.
+			if incremental && (params.MaxIncrements > 0 || params.FullSnapshotAfter > 0) {
+				forceFull, err := s.backupChainExhausted(ctx, params, sourceVolId)
+				if err != nil {
+					return nil, err
+				}
+
+				if forceFull {
+					log.WithField("resource", sourceVolId).Info("incremental backup chain limit reached, forcing a full backup")
+
+					incremental = false
+				}
+			}
+
 			_, err := s.client.Backup.Create(ctx, params.RemoteName, lapi.BackupCreate{
 				RscName:     sourceVolId,
 				SnapName:    id,
-				Incremental: params.AllowIncremental,
+				Incremental: incremental,
 			})
 			if err != nil {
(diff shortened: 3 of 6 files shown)


WanzenBug (approved):

WanzenBug:
Thanks!

PR #450, opened 2026-06-09
Improve test stability

WanzenBug:
The sanity suite started the driver with `go driver.Run()`, which created the unix socket inside the goroutine. On loaded CI runners the test could dial the socket before it was bound; gRPC then backed off and the next retry landed past the suite's 60s connect deadline, failing with "Connection timed out".
    
Split Run() into a package-level Listen() that binds the socket synchronously and a Serve(ctx, listener) method. Callers create the listener first, so the socket accepts connections before Serve runs. Serve takes a context and stops the server gracefully on cancellation, replacing Stop().


rck (approved):
LGTM

PR #452, opened 2026-06-10
fix: verify deployed device sizes when retrying ControllerExpandVolume

mattia-eleuteri:
# fix: verify deployed device sizes when retrying ControllerExpandVolume

## Problem

`ControllerExpand` short-circuits when the volume definition already matches the requested size:

```go
if volumeDefinitionModify.SizeKib == volumeDefinitions[0].SizeKib {
    return nil
}
```

This is unsafe when a previous expand **committed the new volume-definition size but the deployed DRBD device was never actually resized**. The CO (external-resizer) retries `ControllerExpandVolume` with the same arguments, the retry finds a matching volume definition, and reports success — leaving the PVC marked expanded while the block device keeps its old size. For block volumes `NodeExpansionRequired` is `false`, so no node-side step ever catches up.

## How we hit it (production)

LINSTOR/DRBD backend. The deployed DRBD device resize failed on a satellite (a stale DRBD peer left behind by an unrelated `migrate-disk` blocked `drbdadm resize`) **after** LINSTOR had already committed the new volume-definition size. The CSI expand call returned an error, but the volume definition was at the new size. On the retry, the short-circuit above returned success. Result: a 50Gi PVC on a 4Gi DRBD device, silently — the VM booted with a 4Gi disk.

## Scope (please read the discussion)

As @WanzenBug correctly noted, the underlying inconsistency is in **linstor-server**, not the CSI driver: the volume-definition is committed before the deployed device resize is confirmed, and the same-size path of `ModifyVolumeDefinition` is a silent no-op that never re-checks or re-drives the deployed devices. I traced the exact paths in the comment thread.

This PR is therefore a **CSI-side defense-in-depth guard**: it makes the inconsistency visible from the CO (keeps the expand failing instead of silently succeeding) on any LINSTOR version. Happy to have it superseded by / moved to a server-side fix — see comments.

## Fix

When the volume definition already matches the requested size, verify that the deployed volumes actually report the expected usable size (`GetResourceView` filtered on the resource) before returning success. If any device that reports a size is still smaller, return an error so the CO keeps retrying. Devices reporting a usable size of 0 (offline satellite, undeployed device) are skipped, so an unreachable node cannot block expansion forever.

The fresh-expand path is unchanged: `ModifyVolumeDefinition` already surfaces resize errors synchronously.

## Testing

- New unit tests for `ControllerExpand`: fresh expand, successful retry verification, retry with an under-sized device (error, names the node), devices without a reported size (ignored), and shrink rejection.
- `go test ./...` passes; `gofumpt` clean.
- Reproduced on a LINSTOR 1.33 / DRBD 9 cluster (linstor-csi v1.10.6): with this check the retry keeps failing (visible `VolumeResizeFailed` events) until the device is actually resized, instead of silently succeeding.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Deletion of incremental S3 backups no longer fails permanently when a backup
   is still the base of a newer increment. The delete is deferred and retried so
   the chain drains newest-first instead of leaking backups forever.
+- `ControllerExpandVolume` no longer reports success based on the volume
+  definition size alone. A previous expand attempt may update the volume
+  definition while resizing the deployed DRBD devices fails (for example
+  because a replica was not UpToDate at that moment); the retried call then
+  found a matching volume definition and reported success, leaving the block
+  device at its old size while the CO considered the volume expanded. The
+  retry now verifies the usable size reported by the deployed volumes and
+  keeps failing until the devices have actually been resized.
 
 ## [1.11.2] - 2026-05-12
 
--- a/pkg/client/linstor.go
+++ b/pkg/client/linstor.go
@@ -2759,7 +2759,7 @@ func (s *Linstor) ControllerExpand(ctx context.Context, vol *volume.Info) error 
 	}
 
 	if volumeDefinitionModify.SizeKib == volumeDefinitions[0].SizeKib {
-		return nil
+		return s.verifyExpandedSize(ctx, vol.ID, volumeDefinitionModify.SizeKib)
 	}
 
 	if volumeDefinitionModify.SizeKib < volumeDefinitions[0].SizeKib {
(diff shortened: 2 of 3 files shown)


WanzenBug:
Hmm, this seems like a workaround for an issue that is actually within linstor. if linstor rejects the resize, why does it report the updated size in the VD...

mattia-eleuteri:
You're right — thanks for the push. I dug into linstor-server (v1.33.2, the version on our cluster) to confirm, and the root inconsistency is there, not in the CSI driver. Sharing the trace in case it's useful, and I've corrected the PR description (it originally blamed the wrong path — sorry).

There are two distinct paths through `CtrlVlmDfnModifyApiCallHandler.modifyVlmDfnInTransaction`:

**Path A — pre-check rejects, self-healing.** When a replica isn't UpToDate, `allVolumesUpToDate(...)` throws `FAIL_NOT_ALL_UPTODATE` (`"Cannot resize volume, because we have a non-UpToDate DRBD device."`) **before** `ctrlTransactionHelper.commit()`. So the volume-definition is *not* committed, the external-resizer retries, and once everything is UpToDate the resize goes through. No problem, and the CSI driver already handles this fine.

**Path B — satellite resize fails after commit (the silent one).** The pre-check passes, `setVlmDfnSize` + `markVlmDfnResize` get committed (so the VD now reports the new size), then `updateSatellites` runs the actual resize and the DRBD-layer resize fails on a satellite — in our case a stale DRBD peer left by an unrelated `migrate-disk` blocked `drbdadm resize`. The API returns an error, but the volume-definition is already at the new size. That's exactly the *"why does it report the updated size in the VD"* you pointed at.

On the CO's retry, `ModifyVolumeDefinition` lands in the `diffSize == 0` branch, which only emits `WARN_VLMDFN_RESIZE_SAME_SIZE` and returns success **without setting `updateForResize`** — so it never re-checks or re-drives the deployed devices. The CSI driver then sees `vlmdfn size == requested` and returns success too. The PVC ends up marked expanded while the block device is still at its old size (a 50Gi PVC on a 4Gi device, for us).

So the real fix belongs in linstor-server: the `diffSize == 0` path (or the resize flow generally) could verify the deployed `usable_size_kib` and re-drive the resize when the devices are behind the volume-definition, instead of a silent no-op. That would make resize idempotent **and** self-healing, and surface the failure.

This PR is, as you said, a workaround — it only makes the inconsistency visible from the CO side so the PVC isn't marked expanded while the device is smaller. I'm happy to go either way:

- **close this and move it to a linstor-server issue/fix** — I can write up the repro and propose the `diffSize == 0` change there; or
- **keep it as a CSI-side defense-in-depth guard** that works regardless of the LINSTOR version (description now corrected accordingly).

Where would you prefer the fix to live? Happy to do the work in whichever repo.

WanzenBug:
Please fix it upstream, I will close this. Since LINSTOR already does a pre-check, this exact scenario should be unusual enough to not need a full workaround here.

PR #453, opened 2026-06-11
feat: support IO QoS limits via nri-volume-qos

WanzenBug:
StorageClass parameters in the qos.linbit.com/* namespace (rbps, wbps, riops, wiops) are parsed and validated with nri-volume-qos/pkg/meta and stored as auxiliary properties on the LINSTOR resource definition.

On ControllerPublishVolume the limits are read back, tagged with the per-node device path, and returned in the publish context. Kubernetes records that map in VolumeAttachment.status.attachmentMetadata, where the nri-volume-qos plugin reads it to enforce the limits via cgroup io.max.


WanzenBug:
Note: this is useless, unless https://github.com/piraeusdatastore/nri-volume-qos is also deployed.

rck (approved):
LGTM

PR #454, opened 2026-06-12
Run CSI sanity tests against a real LINSTOR cluster

WanzenBug:
Replace the mock-backed CSI sanity test with a `make sanity` target that runs the upstream `csi-sanity` suite against the actual `linstor-csi` binary and a single-node LINSTOR cluster, wired up as a make target and new GitHub Actions job.

The sanity suite in `pkg/driver` ran against an in-memory `MockStorage` backend, so it mostly exercised the mock instead of the driver–LINSTOR interaction. It was also the source of flaky CI runs.


Makefile
```diff
@@ -25,3 +33,37 @@ upload:
...
+	$(SUDO) pvcreate $$(losetup -O NAME -n -j $(SCRATCH_FILE))
+	$(SUDO) vgcreate scratch $$(losetup -O NAME -n -j $(SCRATCH_FILE))
+	$(SUDO) lvcreate --type thin-pool -l 100%FREE --name thin scratch
+# Load dm-snapshot for LINSTOR snapshot support: the satellite cannot
+# modprobe host modules from inside its container.
+	$(SUDO) modprobe dm-snapshot
+	$(CONTAINER_TOOL) run --rm -d --name controller --net host quay.io/piraeusdatastore/piraeus-server:latest startController
+	$(CONTAINER_TOOL) run --rm -d --name satellite --privileged -v /dev:/dev --net host quay.io/piraeusdatastore/piraeus-server:latest startSatellite
+	@echo "Waiting for LINSTOR Controller..."
+	@for i in $$(seq 1 $(CONTROLLER_TIMEOUT)); do if curl -fs -o /dev/null http://localhost:3370/v1/controller/version; then exit 0; fi; sleep 5; done; exit 1
```
rck: I think there is also a `/health` which might be the the more canonical way to test, but really does not matter

rck (approved):
LGTM, just 1 comment (feel free to merge as is)

Makefile
WanzenBug: You are right, thanks!

PR #457, opened 2026-06-17
feat: store filesystem type on the volume definition

WanzenBug:
Set FileSystem/Type and FileSystem/MkfsParams on the volume definition instead of the resource definition. resourceDefinitionToVolume prefers the volume-definition value and falls back to the resource-definition one, so volumes provisioned earlier still resolve their filesystem type.

This lets volumes that share a resource definition use different filesystems, which is a prerequisite for consistency-group volumes.

### DIFF
--- a/pkg/client/linstor.go
+++ b/pkg/client/linstor.go
@@ -263,7 +263,6 @@ func (s *Linstor) resourceDefinitionToVolume(resDef lapi.ResourceDefinitionWithV
 		return nil
 	}
 
-	fsType := resDef.Props[lapiconsts.NamespcFilesystem+"/"+lapiconsts.KeyFsType]
 	props := make(map[string]string)
 
 	for k, v := range resDef.Props {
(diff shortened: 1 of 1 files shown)


pkg/client/linstor.go
```diff
@@ -1744,7 +1751,20 @@ func (s *Linstor) reconcileResourceGroup(ctx context.Context, params *volume.Par
 	return &rg, nil
 }
 
-func (s *Linstor) reconcileResourceDefinition(ctx context.Context, volId, rgName, fsType, mkfsOpts string) (*lapi.ResourceDefinition, error) {
+// fileSystemProps returns the volume-definition properties that drive LINSTOR's automatic filesystem
+// creation, or nil for a raw block volume (no filesystem type).
+func fileSystemProps(fsType, mkfsOpts string) map[string]string {
+	if fsType == "" {
```
rck: maybe a comment when this is the case? at first glance this looks a bit weird that this is even allowed to be `""`.

pkg/client/linstor.go
```diff
@@ -277,6 +277,13 @@ func (s *Linstor) resourceDefinitionToVolume(resDef lapi.ResourceDefinitionWithV
 	for i := range resDef.VolumeDefinitions {
 		vd := &resDef.VolumeDefinitions[i]
 		deviceBytes[int(*vd.VolumeNumber)] = int64(vd.SizeKib) * KiB
+
+		// Prefer the volume definition's filesystem type over the resource-definition-level value.
+		if *vd.VolumeNumber == 0 {
```
rck, line 284: right now this is just some prep step, becuase like this vol 0 wins (if set) and that's it. right?

pkg/client/linstor.go
rck: mv down to the for loop where it might get overwritten (line 276). this keeps "the default" and "with what it can get overwritten" closer together.

rck (changes requested):

pkg/client/linstor.go
WanzenBug: reworked, that this was it's own function was not necessary

pkg/client/linstor.go
WanzenBug: done

pkg/client/linstor.go
WanzenBug: Yes, basically. Updated the comment to be a bit more meaningful.

rck (approved):
LGTM now

PR #458, opened 2026-06-17
refactor: inject Kubernetes client from main

WanzenBug:
Build the Kubernetes clients once in main and pass them to the driver explicitly: a KubeClient option supplies the dynamic client, and ConfigureRWX takes the typed client as an argument.

This enables the creation of a test suite that relies on some k8s API without actually needing to spin up a kubernetes cluster: A fake client can be injected instead.


rck (approved):
yes, guess this works... feel free to merge

PR #459, opened 2026-06-19
refactor: introduce typed volume ID

WanzenBug:
Replace the bare string Info.ID with an embedded volume.ID that carries both the resource name and the volume number, and add ParseVolumeId / String helpers along with an Info.Size() accessor.

Have the attach, detach, publish and expand paths operate on volume.ID and use its volume number instead of assuming the 0th volume.

Update the NFS exporter to address resources by ResourceName, rename the scheduler interface parameters to rdName for clarity, and add tests for the new ID and SnapshotId helpers.

### DIFF
--- a/pkg/client/linstor.go
+++ b/pkg/client/linstor.go
@@ -198,17 +198,18 @@ func (s *Linstor) ListAllWithStatus(ctx context.Context) ([]volume.VolumeStatus,
 	}
 
 	for _, rd := range resDefs {
-		vol := s.resourceDefinitionToVolume(rd)
+		// Listing operates per resource, which maps to the implicit 0th volume of each resource.
+		vol := s.volumeInfoFromResourceDefinition(volume.ID{ResourceName: rd.Name}, rd)
 		if vol != nil {
 			vols = append(vols, volume.VolumeStatus{
 				Info:       *vol,
-				Conditions: ConditionFromResources(resourcesByName[vol.ID]),
+				Conditions: ConditionFromResources(resourcesByName[vol.ResourceName]),
 			})
 		}
 	}
 
 	sort.Slice(vols, func(i, j int) bool {
-		return vols[i].ID < vols[j].ID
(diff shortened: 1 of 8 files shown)


WanzenBug:
FYI, I'm currently running the k8s conformance suite against this change. I don't really expect to have broken anything, but still: touching the whole ID/naming thing is a bit scary.

pkg/client/linstor.go
```diff
@@ -294,7 +294,7 @@ func (s *Linstor) resourceDefinitionToVolume(resDef lapi.ResourceDefinitionWithV
 	}
 
 	return &volume.Info{
-		ID:            resDef.Name,
+		ID:            volume.ID{ResourceName: resDef.Name, VolumeNumber: 0},
```
rck: guess the `0` is intentional here?

pkg/volume/volume.go
```diff
@@ -48,6 +53,48 @@ type Assignment struct {
...
+}
+
+func ParseVolumeId(id string) (ID, error) {
+	if id == "" {
+		return ID{}, fmt.Errorf("empty string is not a valid volume id")
+	}
+
+	volNr := 0
+
+	parts := strings.SplitN(id, "/", 2)
```
rck, line 81: this accepts "/5" as `ID {ResourceName: "", VolumeNumber: 5}`. maybe a guard around `parts[0] == ""`?

rck (commented):
bit scary indeed. LGTM as far as I can tell. mayb don't release it on a Fri afternoon :)

pkg/volume/volume.go
WanzenBug: Fixed. Thanks!

pkg/client/linstor.go
WanzenBug: Kind of. I've reworked it so that the implicit 0 is now only used in the `ListVolumes()` call. At all other points, we already have a proper ID we can use.

PR #461, opened 2026-06-24
clean up attach/detach path

WanzenBug:
We want to clean up any diskless resource we created while attaching. We only want to delete "extra" resources however, and not the tie breaker.

With an update to golinstor, we can pass the "KeepTiebreaker" flag to LINSTOR, so when we delete the diskless, we can be sure it will be taken over as tie breaker, if that is needed.


rck (changes requested):
not a problem with that MR, but the change in golinstor IMO is BS :/. why are the "opts" not optional like all the other "opts" in golinstor? So, NACK from my side, there was no need to break golinstor in the first place. this should be fixed there and then we should depend on a again sane golinstor with optional opts.

WanzenBug:
I also cleaned up the attach path, the workaround with manual resource create is no longer necessary.

rck (approved):
LGTM

PR #462, opened 2026-06-24
feat: carry volume number in snapshot ID

WanzenBug:
Add a VolumeNumber to SnapshotId so the source volume of a snapshot is preserved, mirroring the typed volume ID. The number is encoded as an extra path component (`/<source>/<volume>/<snapshot>`) and falls back to 0, in which case the ID is byte-identical to the previous format.

Populate the number where the source is authoritative (snapshot creation, lookup by ID, lookup by source) and report it via Source().String() as the CSI SourceVolumeId. LINSTOR does not store the CSI-side volume number, so FindSnapsByID recovers it from the requested ID.

### DIFF
--- a/pkg/client/linstor.go
+++ b/pkg/client/linstor.go
@@ -977,6 +977,13 @@ func (s *Linstor) SnapCreate(ctx context.Context, id string, params *volume.Snap
 
 		if s != nil {
 			s.Type = params.Type
+
+			for _, sourceVolId := range sourceVolIds {
+				if sourceVolId.ResourceName == s.SourceName {
+					s.VolumeNumber = sourceVolId.VolumeNumber
+					break
+				}
+			}
 		}
 
 		result = append(result, s)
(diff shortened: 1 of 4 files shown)


rck (commented):
LGTM, AI suggested the following comment:

```diff
diff --git a/pkg/driver/driver.go b/pkg/driver/driver.go
index 95f9d78..0f3446b 100644
--- a/pkg/driver/driver.go
+++ b/pkg/driver/driver.go
@@ -1068,7 +1068,9 @@ func (d *Driver) ListSnapshots(ctx context.Context, req *csi.ListSnapshotsReques

                snapshots = snaps

-               // Regular list of all snapshots.
+               // Regular list of all snapshots. Unlike the by-ID and by-source cases,
+               // there is no requested volume number to recover here, and LINSTOR does
+               // not store the CSI-side number, so snapshots fall back to volume 0.
        default:
                snaps, err := d.linstorClient.ListSnaps(ctx, start, limit)
                if err != nil {
```

also, and I think this is true, this currently is not really multi-volume safe. but this is still some preparation to get there eventually, so most likely fine for now.

WanzenBug:
> also, and I think this is true, this currently is not really multi-volume safe

True, for now we still assume one resource == one volume, and likewise one snapshot == one volume.

This is mostly prep-work so that the eventual support does not need to touch 1000 lines at once :smile: 

I've added a comment, but not where the AI suggested, because it does not make sense there from my perspective.

PR #463, opened 2026-06-25
nfs: add option to keep running on unschedulable nodes

WanzenBug:
We had users that first try to cordon their set of "old" nodes before starting the migration. These users then had the issue that all their NFS exports stopped because the volumes remained on all the unschedulable nodes.

For these users, we want to enable them to keep running the NFS export even on these nodes.

Also fixes an issue where the start-stop-reactor.service could crash without the container getting notified.

### DIFF
--- a/cmd/nfs-helper/start-stop-reactor.go
+++ b/cmd/nfs-helper/start-stop-reactor.go
@@ -15,6 +15,7 @@ import (
 	"k8s.io/apimachinery/pkg/fields"
 	coreinformers "k8s.io/client-go/informers/core/v1"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/utils/env"
 
 	"github.com/piraeusdatastore/linstor-csi/pkg/utils"
 )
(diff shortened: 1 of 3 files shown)


cmd/nfs-helper/start-stop-reactor.go
```diff
@@ -40,12 +41,20 @@ func startStopReactor(ctx context.Context, _ []string) error {
 		options.FieldSelector = fields.OneTermEqualSelector("metadata.name", os.Getenv("NODE_NAME")).String()
 	})
 
+	tolerateUnschedulable := false
+	if v, ok := os.LookupEnv("LINSTOR_CSI_TOLERATE_UNSCHEDULABLE"); ok {
+		tolerateUnschedulable, err = strconv.ParseBool(v)
+		if err != nil {
+			return fmt.Errorf("failed to parse LINSTOR_CSI_TOLERATE_UNSCHEDULABLE='%s' as bool: %w", v, err)
+		}
+	}
```
rck: that has the old question of how one wants to deal with such shell variables. is just setting them already enough? to somehing? to a boolean value? how does that boolean value look like (the string in the given language, "framework", 0/1, "",...). What you have is fine if this is the idiomatic way to deal with this in k8s. is there a helper? is it worth having such a `env2bool` helper in this project?

rck (commented):
LGTM, having a project wide "helper" for handling "boolean" shell variables might be worth it if k8s/csi does not already provide it.

cmd/nfs-helper/start-stop-reactor.go
WanzenBug: Thanks, reused `k8s.io/utils/env` now :smile:

PR #465, opened 2026-07-06
fix: converge parallel CreateVolume requests

WanzenBug:
Concurrent CreateVolume calls that share a resource group, resource definition, or volume definition raced on the "get, then create" reconcilers: the loser got an "already exists" error that failed the whole request. Treat the matching FailExists* return code as success so parallel provisioning converges instead of relying on the external-provisioner retry.

Drop the empty volume group creation in reconcileResourceGroup. The driver creates volume definitions directly and their count varies (one for a normal volume, two for an NFS volume), so a fixed volume-group template is never used; creating it only added another create race.

Use the idiomatic errors.Is(err, lapi.ApiCallRcErr(mask)) throughout instead of the lapi.IsApiCallError helper. ApiCallRcErr is new in golinstor v0.64.0, so bump the dependency.

### DIFF
--- a/pkg/client/linstor.go
+++ b/pkg/client/linstor.go
@@ -1080,7 +1080,7 @@ func (s *Linstor) reconcileBackup(ctx context.Context, id string, params *volume
 func (s *Linstor) backupChainExhausted(ctx context.Context, params *volume.SnapshotParameters, rscName string) (bool, error) {
 	backups, err := s.client.Backup.GetAll(ctx, params.RemoteName, rscName, "")
 	if err != nil {
-		if lapi.IsApiCallError(err, lapiconsts.FailInvldRemoteName) {
+		if errors.Is(err, lapi.ApiCallRcErr(lapiconsts.FailInvldRemoteName)) {
 			// No remote yet -> nothing to base an incremental on.
 			return false, nil
 		}
@@ -1273,7 +1273,7 @@ func (s *Linstor) SnapDelete(ctx context.Context, snap *volume.SnapshotId) error
 		// incremental is "based on" an earlier backup, and LINSTOR refuses to
 		// delete one another increment still depends on, so we need the full set.
 		backups, err := s.client.Backup.GetAll(ctx, snap.Remote, snap.SourceName, "")
-		if err != nil && !lapi.IsApiCallError(err, lapiconsts.FailInvldRemoteName) {
+		if err != nil && !errors.Is(err, lapi.ApiCallRcErr(lapiconsts.FailInvldRemoteName)) {
 			// Ignore errors caused by missing remotes: no remote -> no snapshots to delete.
 			return fmt.Errorf("failed to list backups for snapshot %s: %w", snap.String(), err)
 		}
@@ -1574,6 +1574,11 @@ func (s *Linstor) reconcileSnapshotVolumeDefinitions(ctx context.Context, snapsh
 		restoreConf := lapi.SnapshotRestore{ToResource: targetRD}
 
 		err := s.client.Resources.RestoreVolumeDefinitionSnapshot(ctx, snapshot.SourceName, snapshot.SnapshotName, restoreConf)
+		if errors.Is(err, lapi.ApiCallRcErr(lapiconsts.FailExistsVlmDfn)) {
+			// A concurrent restore already created them; nothing left to do.
+			return nil
+		}
+
 		if err != nil {
 			return fmt.Errorf("could not restore volume definitions: %w", err)
 		}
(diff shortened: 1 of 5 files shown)


pkg/client/linstor.go
```diff
@@ -1573,8 +1573,9 @@ func (s *Linstor) reconcileSnapshotVolumeDefinitions(ctx context.Context, snapsh
 
 		restoreConf := lapi.SnapshotRestore{ToResource: targetRD}
 
+		// A concurrent restore may have created them first.
 		err := s.client.Resources.RestoreVolumeDefinitionSnapshot(ctx, snapshot.SourceName, snapshot.SnapshotName, restoreConf)
-		if err != nil {
+		if err != nil && !errors.Is(err, lapi.ApiCallRcErr(lapiconsts.FailExistsVlmDfn)) {
```
rck: IMO that makes it hard to read and is error prone if code gets added later... maybe make that an

```go
if errors.Is(err, lapi.ApiCallRcErr(lapiconsts.FailExistsVlmDfn) {
    return nil
}
if err != nil {
    bla
}
```

pkg/client/linstor.go
```diff
@@ -1634,8 +1635,9 @@ func (s *Linstor) reconcileSnapshotResources(ctx context.Context, snapshot *volu
 
 	restoreConf := lapi.SnapshotRestore{ToResource: targetRD, Nodes: []string{selectedNode}}
 
+	// A concurrent restore may have created the resource first.
 	err = s.client.Resources.RestoreSnapshot(ctx, snapshot.SourceName, snapshot.SnapshotName, restoreConf)
-	if err != nil {
+	if err != nil && !errors.Is(err, lapi.ApiCallRcErr(lapiconsts.FailExistsRsc)) {
```
rck: same as above

pkg/client/linstor.go
```diff
@@ -1273,7 +1273,7 @@ func (s *Linstor) SnapDelete(ctx context.Context, snap *volume.SnapshotId) error
 		// incremental is "based on" an earlier backup, and LINSTOR refuses to
 		// delete one another increment still depends on, so we need the full set.
 		backups, err := s.client.Backup.GetAll(ctx, snap.Remote, snap.SourceName, "")
-		if err != nil && !lapi.IsApiCallError(err, lapiconsts.FailInvldRemoteName) {
+		if err != nil && !errors.Is(err, lapi.ApiCallRcErr(lapiconsts.FailInvldRemoteName)) {
```
rck: maybe make that more explicit, see comment below where I noticed that for the first time later.

pkg/client/linstor.go
```diff
@@ -1661,19 +1663,18 @@ func (s *Linstor) reconcileResourceGroup(ctx context.Context, params *volume.Par
...
 			SelectFilter: lapi.AutoSelectFilter{},
 		}
-		volumeGroup := lapi.VolumeGroup{}
-		if err := s.client.ResourceGroups.Create(ctx, resourceGroup); err != nil {
-			return nil, err
-		}
-		if err := s.client.ResourceGroups.CreateVolumeGroup(ctx, rgName, volumeGroup); err != nil {
+
+		// A concurrent CreateVolume may have created it first.
+		if err := s.client.ResourceGroups.Create(ctx, resourceGroup); err != nil && !errors.Is(err, lapi.ApiCallRcErr(lapiconsts.FailExistsRscGrp)) {
```
rck: same as above

pkg/client/linstor.go
```diff
@@ -1719,8 +1720,9 @@ func (s *Linstor) reconcileResourceDefinition(ctx context.Context, rdName, rgNam
 			},
 		}
 
+		// A concurrent CreateVolume may have created it first.
 		err = s.client.ResourceDefinitions.Create(ctx, rdCreate)
-		if err != nil {
+		if err != nil && !errors.Is(err, lapi.ApiCallRcErr(lapiconsts.FailExistsRscDfn)) {
```
rck: same as above

rck (changes requested):
I think these new checks should be written differently. not attached to some `err != nil` but as separate "if this happens it is fine and we are done (`return nil`).

pkg/client/linstor.go
WanzenBug: did that where possible. At some code points, we do not `return ...` yet. Those are now an explicit branch that sets `err = nil`. Not sure I like this TBH.

rck: yeah, resetting `err = nil` is ugly, sorry if my comments included such location. This was only intended for locations where we do return, the rest should be kept as it was

WanzenBug: Done

rck (approved):
LGTM now

PR #469, opened 2026-07-07
Consistency groups

WanzenBug:
Adds support for consistency groups: PVCs that share the linstor.csi.linbit.com/consistency-group label are placed as separate volume numbers of a single LINSTOR resource. That makes a VolumeGroupSnapshot
  across them one crash-consistent snapshot — e.g. all disks of a KubeVirt VM — and restore is the mirror: each member reclaims the volume number recorded in its snapshot.

  Gated behind --enable-consistency-groups. Volumes without the label behave exactly as before.


WanzenBug:
@rck This is a rather large change. I am still working on the final commits, I'm not quite happy with some of the code, but everything up to (and including) the "[feat: enumerate consistency-group members on list](https://github.com/piraeusdatastore/linstor-csi/pull/469/changes/982f392bebdb199aa7bc72acc9fcbc566522b8ab)" commit should be pretty much good to go. 

I'll ping you once the other commits are ready.

Makefile
```diff
@@ -42,28 +42,66 @@ linstor-csi:
...
 
+# CLUSTER_UP / CLUSTER_DOWN are the shared cluster fixture for the sanity suites: a loopback LVM-thin pool
+# plus a real LINSTOR controller+satellite (piraeus-server containers) with a node and storage pool. They are
+# kept as variables (not recursive make targets) so the suites can run them inline under a shell EXIT trap
+# without dragging $(MAKE) into a recipe line — a line containing $(MAKE) would execute even under `make -n`.
+#
+# CLUSTER_UP is &&-chained: it stops at the first failure. CLUSTER_DOWN is ;-chained: every step runs even
+# when the cluster only partially came up (a missing LV / VG / PV / loop device / container is fine, we just
+# want it gone), making teardown safe to run after a partial or failed setup.
+CLUSTER_UP = \
```
rck, line 53: pretty ugly. could be a target I guess, where you have the `&&` for free. but if you need it "in the middle" of another target... what in Makefiles is not ugly? just complaining for making me read that early in the morning :).

pkg/client/linstor.go
```diff
@@ -498,8 +498,57 @@ func (s *Linstor) Delete(ctx context.Context, id volume.ID) error {
...
+		if vds[i].VolumeNumber == nil || vds[i].Props[linstor.PropertyCSIVolumeName] == "" {
+			continue
+		}
+
+		if int(*vds[i].VolumeNumber) != id.VolumeNumber {
+			otherMembers++
+		}
+	}
+
+	if otherMembers > 0 {
```
rck: I'd reshuffle that code. claude is good at keeping context, humans are not, return early if possible and keep a happy path without spaghetties on the way:

```go
if otherMembers == 0 {
    return s.deleteResource(ctx, id.ResourceName)
}

// handle member > 0 case:
// the rest even if it duplicates that particular delteResource call
```

pkg/client/linstor.go
```diff
@@ -197,14 +197,30 @@ func (s *Linstor) ListAllWithStatus(ctx context.Context) ([]volume.VolumeStatus,
...
+	for i := range resDefs {
+		rd := resDefs[i]
+
+		// The volume condition is resource-level and shared by every member of a consistency group.
+		cond := ConditionFromResources(resourcesByName[rd.Name])
+
+		isConsistencyGroup := false
+
+		for _, vd := range rd.VolumeDefinitions {
+			if vd.VolumeNumber != nil && vd.Props[linstor.PropertyCSIVolumeName] != "" {
```
rck: I think I saw that "volume number and props"-check several times now? maybe some sort of helper would be nice.

rck (commented):
so far so good, just some minor comments. reviewed until the "enumerate" commit

Makefile
WanzenBug: I originally had that, but that caused some other problems... not sure what they where, so claude came up with this :sweat:

pkg/client/linstor.go
WanzenBug: Done. First use of a go iterator I ever had :smile:

WanzenBug:
Now I think we are at a reviewable state for the rest of the commits, beginning with "deduplicate snapshot..." stuff

pkg/client/linstor.go
```diff
@@ -1977,16 +1977,18 @@ func (s *Linstor) findInProgressOrLegacySnaps(ctx context.Context, id string) ([
...
+		csiSnaps, err := linstorSnapshotToCSISnapshots(&lsnaps[i])
 		if err != nil {
 			return nil, err
 		}
 
-		if snap.Remote != "" {
-			detectedRemotes = true
+		for i := range csiSnaps {
+			if csiSnaps[i].Remote != "" {
+				detectedRemotes = true
```
rck: at least `break`. maybe (also) if-guard the whole loop?

pkg/client/linstor.go
```diff
@@ -503,6 +503,154 @@ func (s *Linstor) Clone(ctx context.Context, vol, src *volume.Info, params *volu
...
+		maps.Copy(props, fileSystemProps(fsType, mkfsOpts))
+
+		err = s.client.ResourceDefinitions.CreateVolumeDefinition(ctx, resource, lapi.VolumeDefinitionCreate{
+			VolumeDefinition: lapi.VolumeDefinition{
+				SizeKib: sizeKib,
+				Props:   props,
+			},
+		})
+		if err != nil {
+			return 0, err
```
rck: using something on `err != nil` is bad behavior, but maybe even return `-1, err` in the error cases to show that this really really isn't a valid number? "nice", but feel free to ignore.

pkg/driver/driver.go
```diff
@@ -635,6 +691,116 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
...
+	if err != nil {
+		return status.Errorf(codes.Internal, "CreateVolume failed for %s: %v", req.GetName(), err)
+	}
+
+	if len(snaps) == 0 {
+		return status.Errorf(codes.NotFound,
+			"CreateVolume failed for %s: snapshot '%s' not found in storage backend", req.GetName(), snapshotID)
+	}
+
+	if len(snaps) > 1 {
```
rck: maybe combine with an `else if` with the previous `== 0` check to "group" these length checks, and maybe use `>=2` to make it a bit more readable.

rck (commented):
hm, only minor things from my side. big change and hard to reason about, but that is just the nature of such big changes I guess.

certainly something not part of this MR, but I'm always a bit surprised when reviewing these changes that there are sooo many `!= ""` checks, always feels that 90% of all types are just strings. I really wonder if some "real" types and helpers would improve anything 🤷

pkg/client/linstor.go
WanzenBug: Replaced with `slices.ContainsFunc`.

WanzenBug:
> certainly something not part of this MR, but I'm always a bit surprised when reviewing these changes that there are sooo many != "" checks, always feels that 90% of all types are just strings. I really wonder if some "real" types and helpers would improve anything 🤷

In rust, I would put `Option<...>` around like 90% of all fields, and most of those would be other enums. No such luck in go :shrug:

pkg/client/linstor.go
rck: still iterates too often because of the outer loop without an `if !detectedRemotes`, but yeah, not that important

rck:
feel free to merge when you think this is ready :boom:

pkg/client/linstor.go
WanzenBug: There are side effects in the outer loop, so we don't want to abort that. But `detectedRemotes` was still buggy, because the last run now overwrote all other checks. In practise, this should be irrelevant, as there the remote should either be set on all or non.

rck: I meant guarding the inner loop, not aborting the outer one :)

PR #474, opened 2026-07-22
Rework Kubernetes client injection

WanzenBug:
Use typed Kubernetes clients instead of the "dynamic" client.

While at it, make it so that all extras that require kubeclients (VolumeSnapshot handling, Consistency Groups, RWX + validation), directly take the clients they depend on as arguments, and are "opt-in" on the code side.

In the case of RWX validation, we keep it opt-out on the CLI side for compatibility. We now also always log a warning if an option is enabled but no kubernetes client could be build.


WanzenBug:
This also replaces https://github.com/piraeusdatastore/linstor-csi/pull/445, as the contributor seems to have abandoned the PR.

rck (approved):
LGTM, feel free to merge

PR #481, opened 2026-08-17
CSI: implement new volume health RPCs

WanzenBug:
CSI spec 1.13 removed the `VOLUME_CONDITION` mechanism; this implements its replacement, the new alpha volume health APIs.

## Notes for review
- Node-side health (`NodeGetVolumeHealth`, `NodeGetStorageHealth`) is intentionally not implemented, matching the previous controller-only condition reporting.

### DIFF
--- a/pkg/client/linstor.go
+++ b/pkg/client/linstor.go
@@ -177,9 +177,9 @@ func LogLevel(s string) func(*Linstor) error {
 	}
 }
 
-// ListAll returns a sorted list of volume and their status.
-func (s *Linstor) ListAll(ctx context.Context) ([]volume.Info, error) {
-	var vols []volume.Info
+// ListAll returns a sorted list of volumes and status.
+func (s *Linstor) ListAll(ctx context.Context) ([]volume.Status, error) {
+	var vols []volume.Status
 
 	resourcesByName := make(map[string][]lapi.ResourceWithVolumes)
 
@@ -200,6 +200,9 @@ func (s *Linstor) ListAll(ctx context.Context) ([]volume.Info, error) {
 	for i := range resDefs {
 		rd := resDefs[i]
 
+		// The volume health is resource-level and shared by every member of a consistency group.
+		healthIssue := HealthIssueFromResource(resourcesByName[rd.Name])
+
 		isConsistencyGroup := false
 
 		for vn := range util.ConsistencyGroupVolumes(rd.VolumeDefinitions...) {
@@ -207,14 +210,14 @@ func (s *Linstor) ListAll(ctx context.Context) ([]volume.Info, error) {
 
 			vol := s.volumeInfoFromResourceDefinition(volume.ID{ResourceName: rd.Name, VolumeNumber: vn}, rd)
 			if vol != nil {
-				vols = append(vols, *vol)
+				vols = append(vols, volume.Status{Info: *vol, HealthIssue: healthIssue})
 			}
 		}
 
 		if !isConsistencyGroup {
 			vol := s.volumeInfoFromResourceDefinition(volume.ID{ResourceName: rd.Name}, rd)
 			if vol != nil {
-				vols = append(vols, *vol)
+				vols = append(vols, volume.Status{Info: *vol, HealthIssue: healthIssue})
 			}
 		}
 	}
@@ -2530,6 +2533,20 @@ func (s *Linstor) FindAssignmentOnNode(ctx context.Context, id volume.ID, node s
 	return va, nil
 }
 
+// GetVolumeHealthIssue returns the health issue of the given volume, if any.
+func (s *Linstor) GetVolumeHealthIssue(ctx context.Context, id volume.ID) (*volume.HealthIssue, error) {
+	s.log.WithFields(logrus.Fields{
+		"volume": id,
+	}).Debug("getting resource health")
+
+	ress, err := s.client.Resources.GetResourceView(ctx, &lapi.ListOpts{Resource: []string{id.ResourceName}})
+	if err != nil {
+		return nil, fmt.Errorf("failed to list resources for '%s': %w", id.ResourceName, err)
+	}
+
+	return HealthIssueFromResource(ress), nil
+}
+
 // SetSnapshotType tries to set the snapshot type based on the remote name.
 // If there is no remote name, it's a local snapshot, if the remote name is an S3 remote, it's an S3 snapshot, etc...
 // If the remote is unknown, we treat it as a local snapshot.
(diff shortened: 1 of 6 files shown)


pkg/client/linstor.go
```diff
@@ -2608,6 +2625,46 @@ func GetSnapshotRemoteAndReadiness(snap *lapi.Snapshot) (string, bool, error) {
...
+
+		if res.State == nil {
+			abnormalNodes = append(abnormalNodes, res.NodeName)
+			continue
+		}
+
+		drbd := util.GetDrbdLayer(res.LayerObject)
+		if drbd != nil && drbd.PromotionScore == 0 {
+			abnormalNodes = append(abnormalNodes, res.NodeName)
+		}
```
rck: guess we can `continue` here as well. for human readability I would include it, especially if then other conditions get added

pkg/client/linstor.go
```diff
@@ -2608,6 +2625,46 @@ func GetSnapshotRemoteAndReadiness(snap *lapi.Snapshot) (string, bool, error) {
 	return "", slices.Contains(snap.Flags, lapiconsts.FlagSuccessful), nil
 }
 
+// HealthFromResources derives the health statuses of a volume from its LINSTOR resources.
+// A resource is abnormal if its state is unknown (satellite offline) or DRBD reports it as not
+// promotable. An empty result means the volume is healthy, abnormal resources degrade the
```
rck: it is not the "empty" result as in "empty list", it is `nil` result? I find that function somehow horrible. IMO it overloads the single return type and after reading it 3 times I find it still hard. maybe multiple return values? or somehow splitting it, don't know, just does not feel right

rck (changes requested):

pkg/client/linstor.go
WanzenBug: You are right, AI produced a convoluted version here. I think I've manged to make it more readable now :smile:

pkg/volume/volume.go
```diff
@@ -200,6 +200,28 @@ type VolumeStats struct {
...
+
+const (
+	// HealthStatusDegraded means the volume works, but some replicas have issues.
+	HealthStatusDegraded HealthStatus = "Degraded"
+	// HealthStatusInaccessible means no healthy replica is left.
+	HealthStatusInaccessible HealthStatus = "Inaccessible"
+)
+
+// Health describes a single health issue of a volume.
+type Health struct {
```
rck: I think the description got it right, "health issue" => `type HealthIssue`, no? then to me a `nil` `HealthIssue` also makes sense (while a `nil` `Health` being a good thing does not really, I personally don't feel very fine with nil health :) )

rck (changes requested):
A lot better code wise, I think the naming is still a bit off?

pkg/volume/volume.go
WanzenBug: Agreed. Renamed to `HealthIssue`, and renamed some of the functions to match.

rck:
# Review

## Overview

Implements the CSI 1.13 alpha volume health APIs (`ControllerGetVolumeHealth`, `ControllerListVolumeHealth`) as a replacement for the removed `VOLUME_CONDITION` mechanism. Health is derived per LINSTOR resource: a replica is abnormal if its state is unknown (satellite offline) or its DRBD promotion score is 0; the volume is then reported as `Degraded` (some abnormal) or `Inaccessible` (all abnormal / no replicas). `ListAll` now returns `volume.Status` (info + optional health issue), the two new capabilities are advertised, and `HealthIssueFromResource` gets solid table-driven tests.

## Bugs

**1. Inverted filter in `ControllerListVolumeHealth` — the RPC reports exactly the wrong volumes** (`pkg/driver/driver.go`)

```go
// Only abnormal volumes are reported.
volumes = slices.DeleteFunc(volumes, func(vol volume.Status) bool {
    return vol.HealthIssue != nil
})
```

`slices.DeleteFunc` removes elements for which the predicate returns **true**, so this deletes every abnormal volume and keeps the healthy ones. The result: the RPC only ever lists healthy volumes, each with a nil `HealthStatuses`. The spec confirms the comment's intent ("The Plugin SHOULD omit volumes with no known adverse health condition when listing all volume health"), so the condition must be `vol.HealthIssue == nil`. As written, the list RPC is a functional no-op; csi-sanity evidently doesn't catch it.

**2. Possible nil-pointer panic in `HealthIssueFromResource`** (`pkg/client/linstor.go`)

`util.GetDrbdLayer(res.LayerObject)` — `LayerObject` is `*lapi.ResourceLayer` (`json:"layer_object,omitempty"` in golinstor v0.65.1) and `GetDrbdLayer` dereferences `layer.Type` without a nil check (`pkg/linstor/util/util.go:102`). A resource returned without layer data would panic inside `ListAll` and both health RPCs. Guard with `res.LayerObject != nil` (or make `GetDrbdLayer` nil-safe).

**3. Negative `starting_token` panics** (`pkg/driver/driver.go`)

`parseAsInt` accepts negative values, and the new code only checks `start > len(volumes)`, so `starting_token: "-1"` reaches `volumes[start:end]` and panics. This mirrors the pre-existing pattern in `ListVolumes`, but since it's being copied into new code, validating `start >= 0` (returning `codes.Aborted` per spec) would be cheap.

## Other issues / suggestions

- **`HealthIssueFromResource` mutates its argument.** `slices.DeleteFunc` shrinks and zeroes the caller's slice — in `ListAll` that's the `resourcesByName[rd.Name]` map entry. Harmless today because each entry is consumed once, but it's a latent footgun for an exported helper. Filtering into a fresh slice (or at least documenting the mutation) would be safer.
- **Test gap on the driver side.** There are good table tests for the health derivation, but none for `ControllerListVolumeHealth`, `ControllerGetVolumeHealth`, or `csiHealthStatus` — a single test asserting that an unhealthy volume appears in the list response would have caught bug 1.
- **Unrelated change:** the removal of the exported `CreatedForTemporaryDisklessAttach` constant from `pkg/linstor/const.go` is a drive-by cleanup with no connection to volume health. Fine to remove if unused, but it belongs in its own commit (and it's part of a public package, so external consumers could notice).
- **`volume.Status` embeds `*HealthIssue`** with no doc comment. A named field (`HealthIssue *HealthIssue`) would read more clearly than an embedded pointer type, avoids surprising field promotion, and the exported type should carry a doc comment per the file's conventions.
- **Pagination consistency:** tokens index into the filtered abnormal-volume list, which changes between calls, so pages can skip or duplicate entries. The spec tolerates this ("results may be inconsistent across calls"), so this is informational — same behavior as `ListVolumes`.
- Minor: `ControllerListVolumeHealth` returns `codes.Internal` when `ListAll` fails while `ListVolumes` returns `codes.Aborted` for the same failure. `Internal` is arguably the better code; just noting the inconsistency.

## What looks good

- Health semantics are well thought out: using `PromotionScore != 0` instead of `may_promote` correctly treats an in-use primary (score > 0, `may_promote: false`) as healthy, and the tie-breaker/diskless cases are covered by tests with realistic API payloads.
- Hoisting the resource-level health above the consistency-group loop in `ListAll` (shared by all group members) is correct and documented.
- `ControllerGetVolumeHealth` handles the missing-ID (`InvalidArgument`), not-found (`NotFound`), and empty-`health_statuses`-means-healthy cases per spec.
- CHANGELOG entry, capability advertisement, and error wrapping all follow existing project conventions.

**Verdict:** the derivation logic is sound, but bug 1 must be fixed before merge — as written, `ControllerListVolumeHealth` never reports an unhealthy volume — ideally together with a driver-level test that would have caught it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

WanzenBug:
Thanks for the thorough review! All three bugs should be fixed with the latest push:

  1. Inverted filter: Good catch. The condition is flipped now, so ControllerListVolumeHealth only reports volumes with a health issue. csi-sanity does exercise the new RPCs, but only against a healthy cluster,
     so it can't tell the difference between "correctly omitted" and "wrongly filtered out".
  2. Nil LayerObject: util.GetDrbdLayer is now nil-safe.
  3. Negative starting token: Now rejected with codes.Aborted. Done as a separate commit, since it applies equally to the existing ListVolumes/ListSnapshots pagination.

  From the other suggestions:

  - HealthIssueFromResource no longer mutates its argument (slices.Clone before filtering).
  - Restored the CreatedForTemporaryDisklessAttach constant. It is unused by the driver itself, but removing it does not belong in this PR.
  - volume.Status now has a named, documented HealthIssue *HealthIssue field instead of the embedded pointer, matching the Health → HealthIssue rename from the earlier round.
  - ListVolumes now also returns codes.Internal when listing fails, so the two RPCs are consistent.
  - Pagination tokens indexing into the filtered list: leaving as is. Same behavior as ListVolumes, and tolerated by the spec.

  On the driver-level tests: agreed that one would have caught the filter bug. The driver is exercised by the external CSI test suite, but that only tests against a healthy cluster. Currently deemed to be too much effort to have yet-another "fake" linstor test.

rck (approved):
LGTM now

PR #484, opened 2026-08-25
Use go tool for go:generate

WanzenBug:
This lets us update the tools automatically using dependabot.

Also work on the mockery deprecation warnings, preparing for an eventual switch to v3.


rck (approved):
LGTM. Verified locally: `go generate ./...` reproduces the committed output byte-for-byte, build and tests pass, and `go mod tidy` is a no-op.

One suggestion (non-blocking): consider adding a generated-code freshness check to CI, e.g. `go generate ./... && git diff --exit-code`. It matters more after this PR, since dependabot will now bump mockery/enumer without regenerating, so the committed mocks could silently drift from what the pinned tool version would produce. `disable-version-string` hides the header churn but not behavioral changes in the generator.

PR #487, opened 2026-08-28
switch tests to typed mock expecters

WanzenBug:
Setting with-expecter matches the mockery v3 default and silences the last deprecation warning. Replace string-based On() calls and hand-built ExpectedCalls slices with the generated EXPECT() API, so expectations are checked against the mocked interface at compile time.


rck (approved):
LGTM

PR #489, opened 2026-08-28
use new LINSTOR 1.35 APIs

WanzenBug:
Use the APIs introduced with LINSTOR 1.35:

* ControllerPublishVolume always invokes MakeAvailable, letting LINSTOR manage allow-two-primaries for RWX volumes via auto-manage-dual-primary instead of toggling the property manually.
* ControllerUnpublishVolume reverts the attachment with UnmakeAvailable.
* Volume deletion truncates the resource definition in one server-side operation, replacing the diskless-first delete loop and the balancer workaround.
* Snapshot deletion removes the empty resource definition together with the last snapshot, replacing the client-side check-then-delete.
* RWX block volumes no longer require a DRBD layer when all configured storage pools are backed by shared storage.
* Fail at startup when the controller does not provide REST API 1.29 (LINSTOR 1.35).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- RWX block volumes no longer require a DRBD layer when all configured storage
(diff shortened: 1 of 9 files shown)


rck (approved):
LGTM

PR #491, opened 2026-08-31
install nfs-ganesha from backports

### LINKED ISSUE #490: piraeus-csi-nfs-server ships NFS-Ganesha 6.5, which breaks RWX volumes on Linux 6.19+ clients
We kept getting Remote I/O error from plain ls on RWX volumes — roughly one in 8000 listings, far worse under load. Took a while to track down.

It's Ganesha 6.5 in piraeus-csi-nfs-server. It answers GET_DIR_DELEGATION with OP_ILLEGAL instead of NFS4ERR_NOTSUPP, which kills the whole COMPOUND, and the client hands EREMOTEIO up to userspace:

COMPOUND: {SEQUENCE, PUTFH, GET_DIR_DELEGATION, GETATTR}
Status of OP_GET_DIR_DELEGATION in position 2 = NFS4ERR_OP_ILLEGAL
That's nfs-ganesha#1385, fixed in 98eb77c4f — but not backported to V6-stable, so 6.5 stays broken. Recent kernels (ours is 7.0) request directory delegations by default, so anyone on one runs into it.

We rebuilt the image with nfs-ganesha from trixie-backports (9.14) and changed nothing else: 0 failures in 28000 listings, server replies NFS4ERR_NOTSUPP.

Would you take a one-liner in nfs/Dockerfile to pull ganesha from ${VERSION_CODENAME}-backports?

Workaround meanwhile is nfsv4.directory_delegations=N on every client node — StorageClass mountOptions doesn't help, the driver appends its own vers=4 and mount.nfs refuses two version options.

munin92:
Fixes #490.

`nfs/Dockerfile`: adds trixie-backports and installs `nfs-ganesha` + `nfs-ganesha-vfs` from there (9.14 instead of 6.5), in a second apt call so the LINBIT packages keep coming from trixie.

Built for amd64: ganesha reports V9.14, and the repo's `default-config.tmpl` parses and starts unchanged. `9.14-1~bpo13+1` exists for both amd64 and arm64.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The driver now requires LINSTOR 1.35 or newer and refuses to start against
   older controllers.
 
+### Fixed
+
+- The NFS server image installs NFS-Ganesha from Debian backports (9.14 instead
+  of 6.5). Ganesha 6.5 answers `GET_DIR_DELEGATION` with `OP_ILLEGAL` rather
+  than `NFS4ERR_NOTSUPP`, which fails the whole COMPOUND and reaches userspace
+  as `EREMOTEIO` on RWX volumes. Linux 6.19 and newer request directory
+  delegations by default, so every client on such a kernel was affected.
+
 ## [1.12.0] - 2026-07-24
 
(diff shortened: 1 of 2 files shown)


WanzenBug:
Thanks!

PR #492, opened 2026-08-31
Switch to mockery v3

WanzenBug:
Switches mock generation from mockery v2 to v3 (v3.7.4), with the minimal set of changes:

- Replace the `vektra/mockery/v2` Go tool dependency with `v3`; convert `pkg/client/.mockery.yaml` to the v3 schema (`template: testify`, `pkgname`/`structname` instead of `outpkg`/`mockname`). The v2-only knobs (`with-expecter`, `disable-version-string`, `issue-845-fix`, `resolve-type-alias`) are gone: expecters are always generated and no version string is emitted.
- Regenerate all mocks from scratch. This drops three files with no counterpart in the current golinstor: `Option` (v3 mocks interfaces only, not function types) and two stale `OneOf*` mocks whose types no longer exist in golinstor v0.65.1. Nothing referenced them.
- Adapt test expectations to v3's variadic handling: v3 mocks pass variadic arguments to `Called()` as a single slice instead of unrolling them, so expectations on variadic methods like `Nodes.GetAll` now match the whole opts slice (and omit the variadic position entirely for calls passing no opts).
- Add `//go:generate rm -rf mocks` before the mockery invocation, so mocks for removed interfaces no longer linger — mockery has no built-in option to prune stale files.


rck (approved):
LGTM

PR #495, opened 2026-09-02
remove dead code

WanzenBug:
Fix the results of running `go run golang.org/x/tools/cmd/deadcode@latest ./...` and remove an unused constant.


rck (approved):
Verified: build, vet, and tests pass on the PR head; no dangling references or orphaned imports; `deadcode ./...` now reports nothing. LGTM.

PR #496, opened 2026-09-07
driver: skip KubeVirt check for storage-only RWX

WanzenBug:
The KubeVirt VM ownership validation guards DRBD's allow-two-primaries. RWX block volumes without a DRBD layer only pass CreateVolume when all storage pools are shared, so there is nothing to guard: concurrent access happens on the shared LV and is coordinated by the users.

Expose this as volume.Info.IsStorageOnly, derived from the resource definition's layer stack, and skip the validator for such volumes in ControllerPublishVolume.

### DIFF
--- a/pkg/client/linstor.go
+++ b/pkg/client/linstor.go
@@ -41,6 +41,7 @@ import (
 	lapiconsts "github.com/LINBIT/golinstor"
 	lapi "github.com/LINBIT/golinstor/client"
 	"github.com/LINBIT/golinstor/clonestatus"
+	"github.com/LINBIT/golinstor/devicelayerkind"
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	"github.com/pborman/uuid"
 	"github.com/sirupsen/logrus"
@@ -310,6 +311,8 @@ func (s *Linstor) volumeInfoFromResourceDefinition(id volume.ID, resDef lapi.Res
 		return nil
 	}
 
+	isStorageOnly := len(resDef.LayerData) == 1 && resDef.LayerData[0].Type == devicelayerkind.Storage
+
 	return &volume.Info{
 		ID:            id,
 		DeviceBytes:   deviceBytes,
@@ -317,6 +320,7 @@ func (s *Linstor) volumeInfoFromResourceDefinition(id volume.ID, resDef lapi.Res
 		FsType:        fsType,
 		Properties:    props,
 		UseQuorum:     resDef.Props["DrbdOptions/Resource/quorum"] != "off",
+		IsStorageOnly: isStorageOnly,
 	}
 }
 
--- a/pkg/driver/driver.go
+++ b/pkg/driver/driver.go
@@ -197,7 +197,8 @@ func ConfigureRWX(cl kubernetes.Interface, namespace, reactorConfigMap string) f
 
 // EnableRWXBlockValidation enables the KubeVirt VM ownership validation for RWX block volumes.
 // When enabled, the driver checks that multiple pods using the same RWX block volume belong to
-// the same VM, guarding against misuse of allow-two-primaries.
+// the same VM, guarding against misuse of allow-two-primaries. Storage-only volumes (shared storage
+// without DRBD) are exempt.
 func EnableRWXBlockValidation(validator *utils.RWXBlockValidator) func(*Driver) error {
 	return func(d *Driver) error {
 		d.rwxBlockValidator = validator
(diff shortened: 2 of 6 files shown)


pkg/client/linstor.go
```diff
@@ -310,13 +311,19 @@ func (s *Linstor) volumeInfoFromResourceDefinition(id volume.ID, resDef lapi.Res
 		return nil
 	}
 
+	// An empty layer stack is unknown rather than storage-only.
+	isStorageOnly := len(resDef.LayerData) > 0 && !slices.ContainsFunc(resDef.LayerData, func(l lapi.ResourceDefinitionLayer) bool {
+		return l.Type != devicelayerkind.Storage
+	})
```
rck: This predicate is stricter than the gate it mirrors on the create side. `CreateVolume`'s shared-storage escape hatch fires whenever the *top* layer isn't DRBD (`driver.go` `params.LayerList[0] != devicelayerkind.Drbd`), so e.g. `luks,storage` or `cache,storage` on shared pools provisions fine — but here such a stack is not "storage-only", so the volume still hits the KubeVirt validator on attach and gets rejected for any non-VM consumer. Fail-closed, so not dangerous, but it contradicts the PR description's rationale ("RWX block volumes without a DRBD layer …").

Suggest deriving the exemption as "no DRBD layer in the stack", which matches the create-side gate exactly and also gets rid of the double negative ("no layer that isn't storage"):

```go
hasDrbd := slices.ContainsFunc(resDef.LayerData, func(l lapi.ResourceDefinitionLayer) bool {
	return l.Type == devicelayerkind.Drbd
})
isStorageOnly := len(resDef.LayerData) > 0 && !hasDrbd
```

(plus a rename, since it's no longer strictly "storage only"). If the stricter predicate is intentional, a comment explaining why LUKS/cache-over-shared stays validated would help the next reader.

Also: this derivation is cheap to pin with a table test on `volumeInfoFromResourceDefinition` — `[STORAGE]`, `[DRBD, STORAGE]`, and empty `LayerData` (the fail-closed case the comment above encodes).

pkg/driver/driver.go
```diff
@@ -849,8 +850,9 @@ func (d *Driver) ControllerPublishVolume(ctx context.Context, req *csi.Controlle
 	// ReadWriteMany block volume
 	rwxBlock := req.VolumeCapability.AccessMode.GetMode() == csi.VolumeCapability_AccessMode_MULTI_NODE_MULTI_WRITER && req.VolumeCapability.GetBlock() != nil
 
-	// Validate RWX block attachment to prevent misuse of allow-two-primaries
-	if rwxBlock && d.rwxBlockValidator != nil {
+	// The validator guards DRBD's allow-two-primaries. A storage-only RWX block volume passed CreateVolume's
+	// shared-storage check, so concurrent access happens on the shared LV and is coordinated by the users.
+	if rwxBlock && !existingVolume.IsStorageOnly && d.rwxBlockValidator != nil {
```
rck: Worth an explicit ack that this is a behavior change beyond the two-primaries concern: the validator also (incidentally) prevented *any* two pods from concurrently attaching a shared-storage RWX block volume from different nodes. After this change, storage-only volumes have no such guard at all — "coordinated by the users" becomes the contract. That's defensible (it matches what other shared-storage CSI drivers do, and RWX block is opt-in), but operators running with validation enabled today may expect it to cover all RWX block volumes.

Related nit: the `-disable-rwx-block-validation` flag help in `cmd/linstor-csi/linstor-csi.go` still reads as if validation covers all RWX block volumes — the `EnableRWXBlockValidation` doc comment got the "storage-only volumes are exempt" clarification, but the user-facing flag text didn't.

rck (commented):
Mechanics check out: `FindByID` embeds the full RD from `ResourceDefinitions.Get`, so `LayerData` is populated on the exact path `ControllerPublishVolume` uses; the derivation is fail-closed for empty/unknown layer stacks; deriving from live LINSTOR state (rather than a stored property) classifies pre-existing volumes correctly; and no new API calls are added.

Two points inline:

1. The publish-time exemption (`stack == [STORAGE]`) is stricter than the create-time gate (`LayerList[0] != drbd` + shared pools), so non-DRBD stacks like `luks,storage` provision but still get validated/rejected on attach — suggest aligning on "no DRBD layer" or documenting the difference.
2. This deliberately drops the concurrent-attach guard for shared-storage volumes entirely; fine, but the CLI flag help text should say so too.

Plus a nit: a small table test on `volumeInfoFromResourceDefinition` would pin the derivation, including the fail-closed empty-`LayerData` case.

pkg/client/linstor.go
WanzenBug: Done: check is now explicitly storage only, as other layers might still be unsafe in this context,i.e. cache on top of shared storage without further coordination is not something we can support.

pkg/driver/driver.go
WanzenBug: RWX was always just for kubevirt workloads. This is an explicit widening because bare LVs in shared storage behave just like any other shared disk.

rck (approved):
LGTM

PR #498, opened 2026-09-09
driver: limit shared-storage RWX to safe layers

WanzenBug:
Without DRBD, LINSTOR activates a shared-storage resource on every node that needs it, so any layer keeping node-local state breaks RWX even for perfectly coordinated consumers: dm-cache and dm-writecache hold dirty or stale blocks on the local cache device, and a flush never pushes them to the shared origin. dm-crypt is a stateless per-sector transform and is as safe as the bare LV.

Only provision RWX block volumes without DRBD for the "storage" and "luks storage" layer lists, and use the same table to decide which volumes skip the KubeVirt validation on publish. This replaces the storage-only marker with volume.Info.SharedStorageSafe.

Follow up from #496

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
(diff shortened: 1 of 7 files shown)


rck (approved):
LGTM

